PR #25, opened 2022-06-23
#14:optimize memtable switch latency

### LINKED ISSUE #14: Import shorten memtable switch latency
improve memtable switch latency with memtables that require large upfront allocation, such as as the hash-table-based memtable

Yuval-Ariel:
With the new speedb memtable, each memtable switch needs to wait
for the allocation and initialization of ~8MiB, which creates
spikes in latency when switching memtables. Add a thread per CF
to allocate new memtables in the background so that they'll be
ready to be used when a switch happens.

Closes #14


isaac-io:
Reverted as part of #76 (in PR #77) due to breakage of unit tests.

Issue #64, opened 2022-07-24
Speedb Artifacts


bosmatt:
- Artifacts from a Speedb build (JAR, Libraries, etc) will include the Speedb name (e.g. libspeedb.a)
- Installation packages (CMake packages, Maven artifactory, apt/yum installers, etc) will include "Speedb"
- The Usage/Version of the library and executables will refer to Speedb

isaac-io:
For the Makefile build artefact names were the only thing that changed, so the build is exactly the same as it was:
* For a static library: `make clean && DEBUG_LEVEL=0 make -j16 static_lib` should create a file named `libspeedb.a`
* For a shared library: `make clean && DEBUG_LEVEL=0 LIB_MODE=shared make -j16 shared_lib` should create files named `libspeedb.so*` with the Speedb version attached (i.e. `libspeedb.so.2.0.0` instead of `librocksdb.so.7.2.2`).
* For the Jar: `make jclean && DEBUG_LEVEL=0 make -j16 rocksdbjavastatic` should create both `libspeedbjni-linux64.so` and `speedbjni-2.0.0-linux64.jar` under `java/target`.
* Otherwise everything should be the same as before, except for the `rocksdb_dump` and `rocksdb_undump` which have been renamed `speedb_dump` and `speedb_undump` respectively.

For the CMake build the package name has been changed from `RocksDB::rocksdb` to `Speedb::speedb` and unfortunately target names also needed to change:
* For a static library: `rm -rf build && mkdir build && cd build && cmake .. -DCMAKE_BUILD_TYPE=Release && make -j16 speedb` should create a file named `libspeedb.a`
* For a shared library: `rm -rf build && mkdir build && cd build && cmake .. -DCMAKE_BUILD_TYPE=Release && make -j16 speedb-shared` should create files named `libspeedb.so*` with the Speedb version attached (i.e. `libspeedb.so.2.0.0` instead of `librocksdb.so.7.2.2`).
* For the Jar: `rm -rf build && mkdir build && cd build && cmake .. -DCMAKE_BUILD_TYPE=Release && make -j16 speedbjni` should create both `libspeedbjni.so` and `speedbjni_classes.jar` under `java`.
* Otherwise here too everything should be the same as before, except for the `rocksdb_dump` and `rocksdb_undump` which have been renamed `speedb_dump` and `speedb_undump` respectively.

[...]

PR #66, opened 2022-07-25
#64: Speedb artifacts

isaac-io:
This PR implements the initial phase of the Speedb artefacts differentiation (#64). That phase deals with renaming the output artefacts from RocksDB to Speedb, as well as replacing any references to RocksDB that are outputted by the code to Speedb.

This includes:
* `librocksdb.{a,so}` -> `libspeedb.{a,so}` (and the SO name was changed from 7.2.2 to 1.5.0)
* `librocksdbjni*.{a,so}` -> `libspeedbjni*.{a,so}` (the same SO name change was made here as well)
* `rocksdbjni*.jar` -> `speedbjni*.jar` (here too, the JAR bears the Speedb version now)

Additionally, the following tools with RocksDB in their name have been renamed:
* `rocksdb_dump` -> `speedb_dump`
* `rocksdb_undump` -> `speedb_undump`

As part of the build changes, all build-time configuration variables were changed from `ROCKSDB_*` to `SPEEDB_*` (e.g. `ROCKSDB_LITE` is now `SPEEDB_LITE`. Note that only the build-time configuration was changed. The macro in the source code is still named `ROCKSDB_LITE`), and the CMake package name is now `Speedb::speedb` instead of `RocksDB::rocksdb`.

Version information that is printed by tools was changed to use the Speedb version information (except for `trace_replay` which uses that information as database version for format compatibility checks).

Additionally, a bunch of scripts were updated to point at the Speedb repository instead of the RocksDB one and to refer to the newly introduced artefact names.

For ease of review, the changes were broken into categories, each in its own commit, the largest of which is unfortunately the first one which deals with build-related changes.

Test Plan: Build all `make` and CMake targets and run `make check` and `make jtest`.


isaac-io:
Updated the PR to reduce the scope of the changes to the artefact names, logging, and version names only. Some of the work (setting the library name based on the project name, removing explicit mentions of RocksDB in status strings, etc.) can theoretically be upstreamable, if RocksDB would be open to accepting something like that.

tools/ldb_tool.cc
```diff
@@ -134,8 +136,7 @@ int LDBCommandRunner::RunCommand(
       PrintHelp(ldb_options, argv[0], /*to_stderr*/ true);
       return 1;
     } else if (std::string(argv[1]) == "--version") {
-      printf("ldb from RocksDB %d.%d.%d\n", ROCKSDB_MAJOR, ROCKSDB_MINOR,
-             ROCKSDB_PATCH);
+      printf("ldb from Speedb %s\n", GetSpeedbVersionAsString().c_str());
```
mrambacher, line 139: Why not use the GetRocksBuildInfoAsString here instead?

tools/db_bench_tool.cc
```diff
@@ -955,12 +956,12 @@ DEFINE_uint64(transaction_lock_timeout, 100,
               " milliseconds before failing a transaction waiting on a lock");
 DEFINE_string(
     options_file, "",
-    "The path to a RocksDB options file.  If specified, then db_bench will "
-    "run with the RocksDB options in the default column family of the "
+    "The path to a database options file.  If specified, then db_bench will "
+    "run with the database options in the default column family of the "
```
mrambacher, line 960: Why not just "The path to an Options file"?

tools/db_bench_tool.cc
```diff
@@ -2697,8 +2697,8 @@ class Benchmark {
 #endif
 
   void PrintEnvironment() {
-    fprintf(stderr, "RocksDB:    version %d.%d\n",
-            kMajorVersion, kMinorVersion);
+    fprintf(stderr, "Speedb:    version %s\n",
+            GetSpeedbVersionAsString(false).c_str());
```
mrambacher, line 2702: Why not just use the GetRocksBuildInfoAsString method?

tools/benchmark.sh
```diff
@@ -259,7 +259,7 @@ function summarize_result {
 
   # if the report TSV (Tab Separate Values) file does not yet exist, create it and write the header row to it
   if [ ! -f "$report" ]; then
-    echo -e "ops_sec\tmb_sec\ttotal_size_gb\tlevel0_size_gb\tsum_gb\twrite_amplification\twrite_mbps\tusec_op\tpercentile_50\tpercentile_75\tpercentile_99\tpercentile_99.9\tpercentile_99.99\tuptime\tstall_time\tstall_percent\ttest_name\ttest_date\trocksdb_version\tjob_id" \
+    echo -e "ops_sec\tmb_sec\ttotal_size_gb\tlevel0_size_gb\tsum_gb\twrite_amplification\twrite_mbps\tusec_op\tpercentile_50\tpercentile_75\tpercentile_99\tpercentile_99.9\tpercentile_99.99\tuptime\tstall_time\tstall_percent\ttest_name\ttest_date\tspeedb_version\tjob_id" \
```
mrambacher, line 262: Just "Version"

fuzz/Makefile
```diff
@@ -38,8 +38,8 @@ else
 # LIB_FUZZING_ENGINE="-fsanitize=fuzzer"
 CC = $(CXX)
 CCFLAGS = $(CXXFLAGS)
-CFLAGS += $(PROTOBUF_CFLAGS) $(PROTOBUF_MUTATOR_CFLAGS) -I$(PROTO_OUT) -I$(ROCKSDB_INCLUDE_DIR) -I$(ROCKSDB_LIB_DIR)
-LDFLAGS += $(PLATFORM_LDFLAGS) $(LIB_FUZZING_ENGINE) $(PROTOBUF_MUTATOR_LDFLAGS) $(PROTOBUF_LDFLAGS) -L$(ROCKSDB_LIB_DIR) -lrocksdb
+CFLAGS += $(PROTOBUF_CFLAGS) $(PROTOBUF_MUTATOR_CFLAGS) -I$(PROTO_OUT) -I$(LIB_INCLUDE_DIR) -I$(LIB_LIB_DIR)
+LDFLAGS += $(PLATFORM_LDFLAGS) $(LIB_FUZZING_ENGINE) $(PROTOBUF_MUTATOR_LDFLAGS) $(PROTOBUF_LDFLAGS) -L$(LIB_LIB_DIR) -l$(LIBNAME:lib%=%)
 endif
```
mrambacher, line 43: I do not think LIBNAME is specified here, is it?

examples/Makefile
```diff
@@ -14,42 +20,42 @@ endif
...
-	$(CXX) $(CXXFLAGS) $@.cc -o$@ ../librocksdb.a -I../include -O2 -std=c++17 $(PLATFORM_LDFLAGS) $(PLATFORM_CXXFLAGS) $(EXEC_LDFLAGS)
+multi_processes_example: static_lib multi_processes_example.cc
+	$(CXX) $(CXXFLAGS) $@.cc -o$@ ../$(LIBNAME).a -I../include -O2 -std=c++17 $(PLATFORM_LDFLAGS) $(PLATFORM_CXXFLAGS) $(EXEC_LDFLAGS)
 
 clean:
 	rm -rf ./simple_example ./column_families_example ./compact_files_example ./compaction_filter_example ./c_simple_example c_simple_example.o ./optimistic_transaction_example ./transaction_example ./options_file_example ./multi_processes_example
 
-librocksdb:
+static_lib:
 	cd .. && $(MAKE) static_lib
```
mrambacher, line 61: You might want to say LIBNAME=$(PROJECTNAME) $(MAKE) static_lib

db_stress_tool/db_stress_gflags.cc
```diff
@@ -56,8 +56,8 @@ DEFINE_double(
 
 DEFINE_string(
     options_file, "",
-    "The path to a RocksDB options file.  If specified, then db_stress will "
-    "run with the RocksDB options in the default column family of the "
+    "The path to a database options file.  If specified, then db_stress will "
+    "run with the database options in the default column family of the "
```
mrambacher, line 60: "to an Options file"

db/db_impl/db_impl.cc
```diff
@@ -4443,7 +4443,7 @@ void DBImpl::EraseThreadStatusDbInfo() const {}
 //
 // A global method that can dump out the build version
 void DumpRocksDBBuildVersion(Logger* log) {
-  ROCKS_LOG_HEADER(log, "SpeeDB version: %s (%s)\n",
+  ROCKS_LOG_HEADER(log, "Speedb version: %s (%s)\n",
```
mrambacher, line 4446: Isn't there a method that gets/prints both already?

cache/cache_bench_tool.cc
```diff
@@ -579,7 +580,8 @@ class CacheBench {
   }
 
   void PrintEnv() const {
-    printf("RocksDB version     : %d.%d\n", kMajorVersion, kMinorVersion);
+    printf("Speedb version     : %s\n",
+           GetSpeedbVersionAsString(false).c_str());
```
mrambacher, line 584: Use the print method.

Makefile
```diff
@@ -2046,18 +2045,18 @@ ifneq ($(origin JNI_LIBC), undefined)
...
-ROCKSDB_JAVA_VERSION ?= $(ROCKSDB_MAJOR).$(ROCKSDB_MINOR).$(ROCKSDB_PATCH)
-ROCKSDB_JAR = rocksdbjni-$(ROCKSDB_JAVA_VERSION)-linux$(ARCH)$(JNI_LIBC_POSTFIX).jar
-ROCKSDB_JAR_ALL = rocksdbjni-$(ROCKSDB_JAVA_VERSION).jar
-ROCKSDB_JAVADOCS_JAR = rocksdbjni-$(ROCKSDB_JAVA_VERSION)-javadoc.jar
-ROCKSDB_SOURCES_JAR = rocksdbjni-$(ROCKSDB_JAVA_VERSION)-sources.jar
+LIB_JAVA_VERSION ?= $(LIB_MAJOR).$(LIB_MINOR).$(LIB_PATCH)
+LIB_JAR = $(PROJECT_NAME)jni-$(LIB_JAVA_VERSION)-linux$(ARCH)$(JNI_LIBC_POSTFIX).jar
+LIB_JAR_ALL = $(PROJECT_NAME)jni-$(LIB_JAVA_VERSION).jar
+LIB_JAVADOCS_JAR = $(PROJECT_NAME)jni-$(LIB_JAVA_VERSION)-javadoc.jar
+LIB_SOURCES_JAR = $(PROJECT_NAME)jni-$(LIB_JAVA_VERSION)-sources.jar
```
mrambacher, line 2059: I am not a huge fan of these LIB_xxx names.  The LIB_ does not really convey much to me.  Seems like we could come up with something better or just drop the lib part?

tools/ldb_tool.cc
isaac-io: `GetRocksBuildInfoAsString()` returns much more than just the version. It also appends all of the build properties, which is unnecessary.

fuzz/Makefile
isaac-io: It is, through the inclusion of `make_config.mk`. The fuzz Makefile is a weird creature. It requires manually building the library with fuzzing enabled and then cd'ing into the fuzz directory and building the fuzzing tools.

db/db_impl/db_impl.cc
isaac-io: Not in this specific format. Changing the format will break tools the grep the log output to find the version (we have internal tools that do just that).

cache/cache_bench_tool.cc
isaac-io: The print method breaks the formatting of the output (alignment of the colon).

tools/db_bench_tool.cc
isaac-io: Because it breaks the output formatting (alignment of the "version ..." string with the subsequently printed lines).

Makefile
isaac-io: Dropping the `LIB_` part is a bit difficult for naming these variables, especially for `LIB_JAVA_VERSION` where it would change the meaning completely. That prefix conveys the meaning that those variables are internal to the library. I can probably append `NAME` where applicable, but it wouldn't make the `LIB_` prefix redundant.

tools/ldb_tool.cc
isaac-io: Done.

tools/db_bench_tool.cc
isaac-io: Done.

examples/Makefile
isaac-io: Good catch. Done.

build_tools/build_detect_platform
```diff
@@ -835,9 +832,9 @@ PLATFORM_CXXFLAGS="$PLATFORM_CXXFLAGS $COMMON_FLAGS"
 
 VALGRIND_VER="$VALGRIND_VER"
 
-ROCKSDB_MAJOR=`build_tools/version.sh major`
-ROCKSDB_MINOR=`build_tools/version.sh minor`
-ROCKSDB_PATCH=`build_tools/version.sh patch`
+LIB_MAJOR=`build_tools/version.sh major`
+LIB_MINOR=`build_tools/version.sh minor`
+LIB_PATCH=`build_tools/version.sh patch`
```
mrambacher, line 837: Nit: I think these would be better named as "MAJOR_VERSION, MINOR_VERSION" etc.  The "LIB" part seems off to me for some reason

mrambacher (approved):
I have not run any of the tests on this code.  If it passes the C++ and Java test suite, good enough for me.

build_tools/build_detect_platform
isaac-io: I renamed them to `VERSION_MAJOR`, `VERSION_MINOR`, and `VERSION_PATCH` to mirror the existing `SHARED_MAJOR` etc. variables.

mrambacher (approved):

PR #68, opened 2022-07-26
initialize mocmemtable members in constructor

ayulas:


isaac-io:
Not needed, as the tests are passing without this change in the latest memtable branch.

PR #83, opened 2022-07-28
71 filter policy uri

### LINKED ISSUE #71: Add construction of filter policy for plugins in tools
While RocksDB supports a plugin infrastructure, it's not used everywhere, and `db_bench`, `db_stress`, and related tools don't allow constructing a custom filter policy from a plugin. This prevents effective testing of filter policies that come from plugins, so we need to add the ability to provide a filter policy URI to the tools.

mrambacher:
Introduced a URI argument that allows the FilterPolicy to be set from the command line to values that are not builtin to RocksDB.  Closes #71

### DIFF
--- a/db_stress_tool/db_stress_test_base.cc
+++ b/db_stress_tool/db_stress_test_base.cc
@@ -30,27 +30,47 @@ namespace ROCKSDB_NAMESPACE {
 namespace {
 
 std::shared_ptr<const FilterPolicy> CreateFilterPolicy() {
-  if (FLAGS_bloom_bits < 0) {
-    return BlockBasedTableOptions().filter_policy;
-  }
-  const FilterPolicy* new_policy;
-  if (FLAGS_use_block_based_filter) {
-    if (FLAGS_ribbon_starting_level < 999) {
-      fprintf(
-          stderr,
-          "Cannot combine use_block_based_filter and ribbon_starting_level\n");
-      exit(1);
-    } else {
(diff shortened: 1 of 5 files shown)


isaac-io (approved):
LGTM.

db_stress_tool/db_stress_test_base.cc
```diff
@@ -34,16 +34,17 @@ std::shared_ptr<const FilterPolicy> CreateFilterPolicy() {
     ConfigOptions config_options;
     std::shared_ptr<const FilterPolicy> policy;
     config_options.ignore_unsupported_options = false;
+    std::string bits_str;
     if (FLAGS_bloom_bits > 0) {
-      fprintf(stderr,
-              "--filter_uri and --bloom_bits are incompatible options\n");
-      exit(1);
+      bits_str = ":" + FormatDoubleParam(FLAGS_bloom_bits);
+      fprintf(stderr, "note: appending --bloom-bits (%f) to --filter-uri\n",
```
mrambacher, line 40: It would be better to print bits-str

mrambacher (commented):
Still don't believe the warning is necessary

db_stress_tool/db_stress_test_base.cc
isaac-io: I used `FLAGS_bloom_bits` directly here instead of `bits_str` in order to avoid having to add pointer arithmetic in order to skip the colon and to avoid having to maintain this in the future if `bits_str` changes.

PR #84, opened 2022-07-28
Devops ci/implement sanity to ci

nathan-bo:


maxb-io (approved):

.github/workflows/ci_pipeline.yml
```diff
@@ -11,11 +11,11 @@ on:
 permissions: write-all
 
 jobs:
-  #Sanity:
-    #uses: speedb-io/speedb/.github/workflows/sanity_check.yml@main
+  Sanity:
+    uses: speedb-io/speedb/.github/workflows/sanity_check.yml@main
```
isaac-io, line 15: Can we not pin to `main` and use the [local repository syntax](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#example-using-an-action-in-the-same-repository-as-the-workflow)?

isaac-io (changes requested):

.github/workflows/ci_pipeline.yml
nathan-bo: yep, done

isaac-io (approved):

isaac-io:
The wrong sanity job now seems to be required for merging after these changes. @maxb-io can you set the right one?

Issue #102, opened 2022-08-02
cmake: allow running the tests in parallel like in the Makefile


isaac-io:
The Makefile supports running tests in parallel using GNU Parallel. CMake runs the tests through [`ctest`](https://cmake.org/cmake/help/latest/manual/ctest.1.html), and is also supposed to run on Windows, so GNU Parallel can't be used.

However, CTest natively supports running tests in parallel with the appropriate environment variable set (`CTEST_PARALLEL_LEVEL`). The problem is that users don't usually have it set, and they might also be used to the `J` environment variable as that's what the Makefile uses.

Add a CMake script in order to be able to set the `CTEST_PARALLEL_LEVEL` (if not set by the user) according to the `J` environment variable if it was set, or fall back to the CPU core count, as the Makefile does.

While at it, allow speeding up the tests on Linux by setting `TEST_TMPDIR` to `/dev/shm` (if exists) according to the Makefile logic (only when the sticky bit is set), otherwise falling back on `/tmp` if exists.

isaac-io:
Tests in the CMake configuration are executed using `ctest`. The default `ctest` output format is two lines for each test, a start line and an end line.

How to test this:

1. Generate a CMake target configuration using your favourite generator (in the example below I'm using the Makefile generator, which is the default on Linux):
   ```
rm -rf build && mkdir build && cd build
cmake ..
make -j$(nproc) check
```
2. When `ctest` starts running you're seeing `$(nproc)` start lines before you start seeing end lines. This is the indication that tests are running in parallel (you'll also see total runtime reduced, but you can only compare before and after).

PR #103, opened 2022-08-02
cmake: allow running the tests in parallel like in the Makefile (#102)

isaac-io:
Allow running tests in parallel by default by setting the CTEST_PARALLEL_LEVEL
environment variable. The functionality is added using a CMake script in
order to determine options when running the tests (rather than when configuring
CMake) in compatibility with the Makefile (in particular -- the `J` and
`TEST_TMPDIR` environment variables).

Test Plan: run the check target using different generators (ninja, make)
to see the improvement in test runtime.

### DIFF
--- a/cmake/CTestRunner.cmake
+++ b/cmake/CTestRunner.cmake
@@ -1,0 +1,113 @@
+# Copyright (C) 2022 Speedb Ltd. All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# 3.12 is needed for FindPython
+cmake_minimum_required(VERSION 3.12)
+
+# Choose the amount of tests to run in parallel if CTEST_PARALLEL_LEVEL wasn't set
+if(NOT DEFINED ENV{CTEST_PARALLEL_LEVEL})
+    # Compatibility with the Makefile: support the `J` environment variable
+    if(DEFINED ENV{J} AND "$ENV{J}" GREATER 0)
+        set(ENV{CTEST_PARALLEL_LEVEL} "$ENV{J}")
+    else()
+        include(ProcessorCount)
+        ProcessorCount(NCPU)
+        if(NOT NCPU EQUAL 0)
+            set(ENV{CTEST_PARALLEL_LEVEL} ${NCPU})
+        endif()
+    endif()
+endif()
+
+# For Makefile compatibility try the following sequence if TEST_TMPDIR isn't set:
+# * Use TMPD if set
+# * Find a suitable base directory and create a temporary directory under it:
+#   * /dev/shm on Linux if exists and has the sticky bit set
+#   * TMPDIR if set and exists
+#   * On Windows use TMP is set and exists
+#   * On Windows use TEMP is set and exists
+#   * /tmp if exists
+if(NOT DEFINED ENV{TEST_TMPDIR})
+    # Use TMPD if set
+    if(DEFINED ENV{TMPD})
+        set(test_dir "$ENV{TMPD}")
+    else()
+        # On Linux, use /dev/shm if the sticky bit is set
+        if("${CMAKE_HOST_SYSTEM_NAME}" STREQUAL "Linux" AND IS_DIRECTORY "/dev/shm")
+            execute_process(COMMAND test -k /dev/shm RESULT_VARIABLE status OUTPUT_QUIET ERROR_QUIET)
+            if(status EQUAL 0)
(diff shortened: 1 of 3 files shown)


cmake/CTestRunner.cmake
```diff
@@ -0,0 +1,59 @@
...
+        if(status EQUAL 0)
+            set(test_dir /dev/shm)
+        endif()
+    endif()
+    # Fall back to /tmp if exists
+    if(NOT DEFINED test_dir AND IS_DIRECTORY /tmp)
+        set(test_dir /tmp)
+    endif()
+    # Create a temporary directory under the test location that we determined
+    if(DEFINED test_dir)
```
mrambacher, line 42: This is not the same as the Makefile. 
- I think tests always create a directory under TEST_TMPDIR of the form rocksdb.XXX.  
- The Makefile does this on all non-Windows machines (not just Linux).
- If TEST_TMPDIR is not set, it should default to TMPDIR (or /tmp)

cmake/CTestRunner.cmake
```diff
@@ -0,0 +1,59 @@
...
+        endif()
+        set(ENV{TEST_TMPDIR} ${tmpdir})
+    endif()
+endif()
+
+if(DEFINED ENV{TEST_TMPDIR})
+    message(STATUS "Running $ENV{CTEST_PARALLEL_LEVEL} tests in parallel in $ENV{TEST_TMPDIR}")
+endif()
+
+# Run all tests, with a timeout of 10 minutes per test and show test output on failure
```
mrambacher, line 58: Can we make the 10 minute timeout a variable/configurable?

cmake/CTestRunner.cmake
isaac-io: * Tests create their directories in the following format: `${TEST_TMPDIR:-/tmp/rocksdbtest-$(id -u)}/<test_name>_<pid>_<tid>`. This is not a safe thing to do when running tests in parallel due to PID and TID reuse (not to mention two tests runs on the same machine).
* Right, but the Makefile is wrong. `/dev/shm` is a Linux thing, not a POSIX thing, and many POSIX compliant OSs simply don't have it, or don't allow it to be used as a general purpose fileystem. I'll allow the rest of the temporary directory creation logic to work for other POSIX systems as well though.
* Will do.

cmake/CTestRunner.cmake
isaac-io: Done.

mrambacher:
I am guessing this parallel test is invoked via J=# make check, but when I do that, I get an error:

 J=20 make check

CMake Warning (dev) at /Users/markrambacher/Documents/GitHub/speedb/cmake/CTestRunner.cmake:19 (set):
  Syntax error in cmake code at

    /Users/markrambacher/Documents/GitHub/speedb/cmake/CTestRunner.cmake:19

  when parsing string

    ${ENV{J}}

  syntax error, unexpected {, expecting } (6)

  Policy CMP0010 is not set: Bad variable reference syntax is an error.  Run

isaac-io:
Oops, my bad (a leftover from earlier experimentation). I pushed a fix for this.

mrambacher (approved):

cmake/CTestRunner.cmake
```diff
@@ -0,0 +1,85 @@
...
+        if("${CMAKE_HOST_SYSTEM_NAME}" STREQUAL "Linux" AND IS_DIRECTORY /dev/shm)
+            execute_process(COMMAND test -k /dev/shm RESULT_VARIABLE status OUTPUT_QUIET ERROR_QUIET)
+            if(status EQUAL 0)
+                set(test_dir /dev/shm)
+            endif()
+        endif()
+        # Use TMPDIR as base if set
+        if(NOT DEFINED test_dir AND IS_DIRECTORY "$ENV{TMPDIR}")
+            set(test_dir $ENV{TMPDIR})
+        endif()
```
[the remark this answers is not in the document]
isaac-io: According to [a Raymond Chen post on the subject](https://devblogs.microsoft.com/oldnewthing/20150417-00/?p=44213), we should probably check both `TMP` and `TEMP` on Windows. I'll add that (though the code in `WinFileSystem::GetTestDirectory()` already checks `TMP` on its own in case `TEST_TMPDIR` isn't set).

cmake/CTestRunner.cmake
```diff
@@ -0,0 +1,85 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# Choose the amount of tests to run in parallel if CTEST_PARALLEL_LEVEL wasn't set
+if(NOT DEFINED ENV{CTEST_PARALLEL_LEVEL})
+    # Compatibility with the Makefile: support the `J` environment variable
+    if(DEFINED ENV{J} AND "$ENV{J}" GREATER 0)
```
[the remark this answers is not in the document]
isaac-io: Technically it doesn't really matter, because if `CTEST_PARALLEL_LEVEL` isn't set the behaviour would be as if `J=1` anyway. It's just easier that way to show that we respect `J` if it's numeric and positive.

mrambacher (approved):
LGTM

mrambacher (approved):

PR #106, opened 2022-08-03
Devops ci/add license verification pipeline

nathan-bo:
added license verification pipeline and implement to ci


.github/workflows/license_verify.yml
```diff
@@ -0,0 +1,38 @@
...
+      - name: Checkout repository
+        uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
+      - name: echo changed files
+        run: |
+          exit_code=0
+          for file in $(echo ${{needs.changedfiles.outputs.changed_files}})
+          do
+            if ! grep -q "Copyright (C) 2022 Speedb Ltd. All rights reserved." "$file"; then
```
rtpro, line 34: @nathan-bo Please note that the year should not be part of the grep, as it would change next year.. :)

rtpro (changes requested):

.github/workflows/license_verify.yml
nathan-bo: fixed, now it will use the current year instead of a set string :)

nathan-bo: Changed it to regex to look for 20** as the year, let me know if you want this changed

maxb-io: @nathan-bo I still see the hardcoded "2022". You can use regex for limiting to year 20XX: 
if ! grep -q "Copyright (C) 20.. Speedb Ltd. All rights reserved." "$file"; then
or catch the current year:
 grep -q  "Copyright (C) `date +%Y` Speedb Ltd. All rights reserved." 
 grep -q  "Copyright (C) $(date +%Y) Speedb Ltd. All rights reserved."

Issue #126, opened 2022-08-18
flush_job: do not roll back memtable flush on CF drop or database shutdown


isaac-io:
In non-atomic flush the flush job code currently rolls back the memtable flush in case the CF was dropped or if the database is shutting down. This is unnecessary since it reactivates the pending flush flag on the relevant CF even though no flush should take place after CF drop or DB shutdown request, and it also causes an issue for the upcoming improvements to the `WriteBufferManager` immutable memory tracking (#113) because it marks the memtable memory as ready for flush again for a short period before finally getting freed when the memtable is dropped during the destruction of the dropped CF or the closing of the DB (which might upset the WBM behaviour because in the future we plan to rely on the amount of memory that isn't already being flushed for triggering flushes rather than on the amount of mutable memory alone as is done today).

Since the rollback only marks the memtable as ready for flush again and clears related flush state, there's no harm in skipping it and just letting the memtable drop in that state. Moreover, in atomic flush those cases do not trigger a rollback for flushes that already completed, so this simply matches the behaviour there.

Issue #113, opened 2022-08-04
WriteBufferManager: accurate memory tracking of immutable memory


isaac-io:
Currently the WriteBufferManager tracks active memory through hooks into the memtable allocation and deallocation logic (using the `AllocTracker`. See [memtable.h](https://github.com/speedb-io/speedb/blob/1c7a33ae5ba2ce036696ba7d5baa4561bb59584a/db/memtable.h) and [allocator.h](https://github.com/speedb-io/speedb/blob/1c7a33ae5ba2ce036696ba7d5baa4561bb59584a/memory/allocator.h)).

This allows it to know how much mutable and immutable memory is currently taken up. However immutable memory isn't necessarily memory that can be released. For example, if a flush has been triggered on the immutable memory, that memory is already marked for deallocation, so triggering another flush there won't do any good.

For this reason, the WriteBufferManager currently does almost nothing with the immutable memory count. The current implementation takes action based on the immutable memory size only if capacity has been reached and half of that capacity is taken up by mutable memory.

We need to be able to know which immutable memory isn't already marked for flush, and for that the WriteBufferManager needs to hook into the flush logic in order to take that into account. Of note here is that memory is only marked for flush after picking memtables for flush (see [db_impl_compaction_flush.cc#L239](https://github.com/speedb-io/speedb/blob/0a7b4eb48581c8fb5a16103306b42b2aef800da5/db/db_impl/db_impl_compaction_flush.cc#L239) for regular flush and [db_impl_compaction_flush.cc#L503](https://github.com/speedb-io/speedb/blob/0a7b4eb48581c8fb5a16103306b42b2aef800da5/db/db_impl/db_impl_compaction_flush.cc#L503) for atomic flush), so there's still a span of time between triggering a flush and the time this memory will be marked as being in progress. An implementation note for when we start taking action based on immutable memory size in the future, is that we might need a way to wait for triggered flushed to pick memtables before we rely on the immutable memory size, but that's out of scope for this issue.

Note that we need to "unmark" the memory that is marked for flush when the flush is done and the immutable memory is released (which can be combined into the existing `FreeMem()` hook), as well as when the flush job is cancelled (such as when a CF is dropped) before memtables have been picked (in which case relying on `FreeMem()` would be the wrong thing to do because it would subtract memory marked for flush even though no memory has been marked yet.

isaac-io:
Depends on #144.

assaf-speedb:
Unit tests, stress and fuzz passed

[...]

PR #127, opened 2022-08-19
flush_job: do not roll back memtable flush on CF drop and DB shutdown (#126)

isaac-io:
Rolling back in these cases is meaningless because the memtables will be
freed shortly afterwards, so making them available for flush again is useless
at best.

Additionally, in atomic flush, memtables that were successfully flushed
aren't rolled back if a CF is dropped or a DB shutdown was requested (the
shutdown case is the more relevant one for the comparison here because it
aborts the flush, whereas the in the CF drop case it's simply ignored at
flush result installation time). In this regard this change simply matches
the behaviour there.

Lastly, this might be needed for the WriteBufferManager changes in #113,
since we plan to use the information about memory that can be flushed in
the future for triggering flushes based on immutable memory as well, and
rolling back flushes causes the memtable memory to become ready for flush
again for a brief period of time until it is dropped, which might wrongly
affect the WBM decisions.

Test plan: run `make check` as well as `make crash_test`.

### DIFF
--- a/db/flush_job.cc
+++ b/db/flush_job.cc
@@ -276,16 +276,12 @@ Status FlushJob::Run(LogsWithPrepTracker* prep_tracker, FileMetaData* file_meta,
     s = WriteLevel0Table();
   }
 
-  if (s.ok() && cfd_->IsDropped()) {
-    s = Status::ColumnFamilyDropped("Column family dropped during compaction");
-  }
-  if ((s.ok() || s.IsColumnFamilyDropped()) &&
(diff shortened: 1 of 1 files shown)


udi-speedb:
@isaac-io: Did make check completed successfully on this branch?

isaac-io:
> @isaac-io: Did make check completed successfully on this branch?

Yes. Both `make check` and `make crash_test` completed successfully.

udi-speedb (approved):

Issue #128, opened 2022-08-19
build: fix Java test depedndencies in the Makefile


isaac-io:
The target dependencies are incorrectly specified for the `test` target in `java/Makefile`. This causes it sometime fail when make is invoked in job server mode (`-j` set a something other than 1). Fix the target dependencies so that the tests are run only after everything was built.

PR #129, opened 2022-08-19
build: fix the java test target dependencies (#128)

isaac-io:
The `test` target in the Java Makefile has an incorrect dependency graph
that could cause tests to start executing before the compilation finished
(or even started) when make is invoked in job server mode.

Fix it by specifying target dependencies such that running the tests depends
on having the tests compiled first.


mrambacher (approved):
LGTM

PR #132, opened 2022-08-23
Accurate memory tracking of immutable memory

udi-speedb:


memtable/write_buffer_manager.cc
```diff
@@ -58,26 +59,20 @@ std::size_t WriteBufferManager::dummy_entries_in_cache_usage() const {
...
+  auto curr_mem_used = memory_used_.fetch_add(mem, std::memory_order_relaxed);
+
   if (cache_res_mgr_ != nullptr) {
-    ReserveMemWithCache(mem);
-  } else if (enabled()) {
-    memory_used_.fetch_add(mem, std::memory_order_relaxed);
-  }
-  if (enabled()) {
-    memory_active_.fetch_add(mem, std::memory_order_relaxed);
+    ReserveMemWithCache(curr_mem_used + mem);
```
isaac-io, line 65: This is an unsafe change. Consider the following scenario with two threads, T1 and T2:

1. T1 Calls `ReserveMem(5)` and gets `curr_mem_used = 0`.
2. T2 Calls `ReserveMem(7)` and gets `curr_mem_used = 5`.
3. T2 Calls `ReserveMemWithCache(12)` (5 + 7).
4. T1 Calls `ReserveMemWithCache(5)` (0 + 5).

That's why the original code does the increment of used memory under mutex, in order to ensure that the cache gets a correct reservation, and the atomic increment was used only if cache reservation isn't configured.

memtable/write_buffer_manager.cc
```diff
@@ -92,29 +87,42 @@ void WriteBufferManager::ReserveMemWithCache(size_t mem) {
 }
 
 void WriteBufferManager::ScheduleFreeMem(size_t mem) {
-  if (enabled()) {
-    memory_active_.fetch_sub(mem, std::memory_order_relaxed);
-  }
+  memory_inactive_.fetch_add(mem, std::memory_order_relaxed);
```
isaac-io, line 90: This should only be done if enabled. No point in doing unnecessary atomic RMW operations.

memtable/write_buffer_manager.cc
```diff
@@ -92,29 +87,42 @@ void WriteBufferManager::ReserveMemWithCache(size_t mem) {
...
+}
+
+void WriteBufferManager::FreeMemBegin(size_t mem) {
+  memory_being_freed_.fetch_add(mem, std::memory_order_relaxed);
```
isaac-io, line 94: Same here: only do this if `enabled()`.

memtable/write_buffer_manager.cc
```diff
@@ -92,29 +87,42 @@ void WriteBufferManager::ReserveMemWithCache(size_t mem) {
...
+}
+
+// Freeing 'mem' bytes was aborted and that memory is no longer in the process
+// of being freed
+void WriteBufferManager::FreeMemAborted(size_t mem) {
+  assert(memory_being_freed_.load(std::memory_order_relaxed) >= mem);
+  memory_being_freed_.fetch_sub(mem, std::memory_order_relaxed);
```
isaac-io, line 101: This assertion is unsafe. You should be doing something along the following lines:

```c++
    const size_t being_freed_before = memory_being_freed_.fetch_sub(mem, std::memory_order_relaxed);
    if (being_freed_before < mem) {
      assert(false);
    }
```

`assert(false)` is used (rather than `assert(being_freed_before >= mem)`) in order to avoid unused variable diagnostic in release builds.

memtable/write_buffer_manager.cc
```diff
@@ -92,29 +87,42 @@ void WriteBufferManager::ReserveMemWithCache(size_t mem) {
...
 }
 
 void WriteBufferManager::FreeMem(size_t mem) {
+  assert(memory_used_.load(std::memory_order_relaxed) >= mem);
+  assert(memory_inactive_.load(std::memory_order_relaxed) >= mem);
+  assert(memory_being_freed_.load(std::memory_order_relaxed) >= mem);
```
isaac-io, line 107: Same here: asserting on atomic variables before doing the modification is unsafe.

memtable/write_buffer_manager.cc
```diff
@@ -92,29 +87,42 @@ void WriteBufferManager::ReserveMemWithCache(size_t mem) {
...
+
+  auto curr_mem_used = memory_used_.fetch_sub(mem, std::memory_order_relaxed);
+  memory_inactive_.fetch_sub(mem, std::memory_order_relaxed);
+  memory_being_freed_.fetch_sub(mem, std::memory_order_relaxed);
```
isaac-io, line 111: Same as above: no point in doing atomic RMW if not `enabled()`.

memtable/write_buffer_manager.cc
```diff
@@ -92,29 +87,42 @@ void WriteBufferManager::ReserveMemWithCache(size_t mem) {
...
+
   if (cache_res_mgr_ != nullptr) {
-    FreeMemWithCache(mem);
-  } else if (enabled()) {
-    memory_used_.fetch_sub(mem, std::memory_order_relaxed);
+    FreeMemWithCache(curr_mem_used - mem);
```
isaac-io, line 114: This change is unsafe, as explained above. With cache reservation enabled we should only be changing the used memory under the cache reservation mutex.

isaac-io (changes requested):

include/rocksdb/write_buffer_manager.h
```diff
@@ -71,7 +71,19 @@ class WriteBufferManager final {
...
+    const size_t inactive = memory_inactive_.load(std::memory_order_acquire);
+    return ((inactive >= total) ? 0 : (total - inactive));
+  }
+
+  // Returns the total inactive memory used by memtables.
+  size_t immmutable_memtable_memory_usage() const {
+    return memory_inactive_.load(std::memory_order_relaxed);
+  }
+
+  // Returns the total memory used by active memtables.
```
isaac-io, line 84: This comment refers to the wrong memory category.

include/rocksdb/write_buffer_manager.h
```diff
@@ -81,9 +93,19 @@ class WriteBufferManager final {
     return buffer_size_.load(std::memory_order_relaxed);
   }
 
+  // Note that the memory_inactive_ and memory_being_freed_ counters
+  // are NOT maintained when the WBM is disabled. In addition, memory_used_ is
+  // maintained only when enabled or cache is provided. Therefore, if switching
+  // from disabled to enabled, these counters will (or may) be invalid or may
+  // wraparound
```
isaac-io, line 100: This seems like an upstream bug, regardless of the changes in this PR. Can we open a separate issue to fix this (i.e. track memory state even when disabled in order to avoid this scenario) and make this PR depend on it?

memtable/alloc_tracker.cc
```diff
@@ -44,10 +45,41 @@ void AllocTracker::DoneAllocating() {
...
+    write_buffer_manager_->FreeMemAborted(
+        bytes_allocated_.load(std::memory_order_relaxed));
+  }
+}
+
 void AllocTracker::FreeMem() {
+  if (freed_) {
+    return;
+  }
+
```
isaac-io, line 74: This check is redundant. The code below already checks for `!freed_` to avoid freeing when not needed.

isaac-io (commented):
I left a few minor comments, but other than that LGTM. Can you also add a test or two to ensure the new functionality doesn't regress in the future?

memtable/alloc_tracker.cc
udi-speedb: If FreeMem() is called twice without this check, the second time it will call FreeMemBegin(), potentially assert(!freed), and do a bunch of unneeded stuff.

isaac-io: I see. Can we rename `is_memory_being_freed_` to `memory_free_started_` and change the semantics so that it's not reset below? (only `FreeMemAborted()` should be the one to reset it)

Alternatively, since this is simply a state machine, we can replace all of these members with a single `state_` member of type `AllocTrackerState` which would be defined in the following way:

```c++
enum class AllocTrackerState {
  kAllocating,
  kDoneAllocating,
  kFreeMemStarted,
  kFreed,
};
```

That way `ReserveMem()` could assert that it's in `kAllocating`, `DoneAllocating()` should assert the same and change the state to `kDoneAllocating`, `FreeMemBegin()` would assert that the state is `kDoneAllocating` and set it to `kFreeMemStarted`, and `FreeMemAborted` can assert that the state is `kFreeMemStarted` and revert it back to `kDoneAllocating`.

Then `FreeMem()` could be defined as follows:

```c++
switch (state_) {
case AllocTrackerState::kAllocating:
  DoneAllocating();
  FALLTHROUGH_INTENDED;
case AllocTrackerState::kDoneAllocating:
  FreeMemBegin();
  FALLTHROUGH_INTENDED;
case AllocTrackerState::kFreeMemStarted:
  ...
  state_ = AllocTrackerState::kFreed;
  FALLTHROUGH_INTENDED;
case AllocTrackerState::kFreed:
  break;
default:
  assert(false);
}
```

include/rocksdb/write_buffer_manager.h
udi-speedb: I have opened [https://github.com/speedb-io/speedb/issues/138](url)
How do we make a PR depend on it and what does it mean practically?

isaac-io: Thanks!
We currently have no means of specifying that a PR depends on an issue or another PR because the automation is still a WIP, so generally making something dependent on another is currently simply adding `Depends on ...` to the PR description and avoiding merging it until the other one is resolved. My thinking was that since that's an upstream bug, it would be easier if we fixed it before making the changes in this PR, so that it'd be easier to upstream. However, thinking about it some more, I don't think that the failure case justifies holding back this PR if you won't have time to open a PR for the other issue and add a regression test.

memtable/write_buffer_manager.cc
```diff
@@ -95,7 +95,7 @@ void WriteBufferManager::ScheduleFreeMem(size_t mem) {
   }
 }
 
-void WriteBufferManager::FreeMemBegin(size_t mem) {
+void WriteBufferManager::FreeMemStarted(size_t mem) {
```
isaac-io, line 98: Why the change in naming style here? `FreeMemBegin()` was following the imperative style of `ReserveMem()` and `ScheduleFreeMem()`, as well as `FreeMem()`. Changing this to `FreeMemStarted()` makes it unnecessarily different.

memtable/alloc_tracker.cc
```diff
@@ -15,81 +15,78 @@
...
+        (state_ == State::kFreeMemStarted)) {
       write_buffer_manager_->FreeMem(
           bytes_allocated_.load(std::memory_order_relaxed));
     } else {
       assert(bytes_allocated_.load(std::memory_order_relaxed) == 0);
     }
-    is_memory_being_freed_ = false;
-    freed_ = true;
   }
+  bytes_allocated_ = 0U;
```
isaac-io, line 89: Do we really need to reset this considering the fact that it's not exposed anywhere? The state machine should work regardless of its value (and it may actually be nice to have that number if we end up wanting to know in the future how much memory was allocated, even after it was freed, for statistics etc.).

memtable/alloc_tracker.cc
```diff
@@ -15,81 +15,78 @@
...
-  if (write_buffer_manager_->enabled() ||
-      write_buffer_manager_->cost_to_cache()) {
+  assert((state_ == State::kNone) || (state_ == State::kAllocating));
+
+  if (ShouldUpdateWriteBufferManager() &&
+      ((state_ == State::kNone) || (state_ == State::kAllocating))) {
     bytes_allocated_.fetch_add(bytes, std::memory_order_relaxed);
     write_buffer_manager_->ReserveMem(bytes);
   }
+  state_ = State::kAllocating;
```
isaac-io, line 31: The AllocTracker starts its life under the allocator and `Allocate()` is called many times. What does the `kNone` state signify to warrant the additional checks and reassignment over and over again?

Also, note that you're resetting the state machine to the allocating state even if the assertion doesn't hold. Since you're checking it anyway (so there's no additional cost to doing it only if the state is the expected one), I'm not sure that it's the right thing to do (the same goes for the rest of the state changes below).

memtable/write_buffer_manager.cc
udi-speedb: The alloc_tracker is not actually freeing the memory, it is notified of the fact that memory is being freed (more like DoneAllocating). Having said that, both are fine IMO.

isaac-io: That's why I only commented about the change in the WBM and not in the AllocTracker. As for `FreeMemBegin()` vs. `BeginFreeMem()`, I'm fine with either.

db/memtable_list.cc
```diff
@@ -701,8 +701,9 @@ void MemTableList::RemoveMemTablesOrRestoreFlags(
                          m->edit_.GetBlobFileAdditions().size(), mem_id);
       }
 
-      m->flush_completed_ = false;
-      m->flush_in_progress_ = false;
+      m->SetFlushInProgress(false);
+      m->SetFlushCompleted(false);
```
isaac-io, line 705: I'm only now noticing that you changed the order of these two lines. Is that because we're depending on the value provided to us in `SetFlushCompleted()`? If so, we could make sure that it's different than `flush_completed_` as well before we take action.

db/memtable_list.cc
```diff
@@ -893,8 +894,8 @@ Status InstallMemtableAtomicFlushResults(
                            edit->GetBlobFileAdditions().size(), mem_id);
         }
 
-        m->SetFlushCompleted(false);
         m->SetFlushInProgress(false);
+        m->SetFlushCompleted(false);
```
isaac-io, line 898: Same here: why the ordering change?

db/memtable_list_test.cc
```diff
@@ -541,313 +541,63 @@ TEST_F(MemTableListTest, GetFromHistoryTest) {
...
       MemTable* mem =
-          new MemTable(cmp, ioptions, *(mutable_cf_options_list.back()), &wb,
-                       kMaxSequenceNumber, cf_id);
+          new MemTable(cmp, ioptions, mutable_cf_options, &wb,
+                       kMaxSequenceNumber, 0 /* column_family_id */);
       mem->SetID(memtable_id++);
       mem->Ref();
 
+      auto new_total_reserved_mem = wb.mutable_memtable_memory_usage();
+      ASSERT_TRUE(!wbm_enabled ||
```
isaac-io, line 593: Can we put it inside an if statement rather than wrapping it in `ASSERT_TRUE`? in case of failure it'll be hard to understand what the expected vs. actual was.

db/memtable_list_test.cc
```diff
@@ -860,100 +610,488 @@ TEST_F(MemTableListTest, AtomicFlusTest) {
...
+      tables_reserved_mem[cf_id].resize(num_tables_per_cf);
+      for (int i = 0; i != num_tables_per_cf; ++i) {
+        MemTable* mem =
+            new MemTable(cmp, ioptions, *(mutable_cf_options_list.back()), &wb,
+                         kMaxSequenceNumber, cf_id);
+        mem->SetID(memtable_id++);
+        mem->Ref();
+
+        auto new_total_reserved_mem = wb.mutable_memtable_memory_usage();
+        ASSERT_TRUE(!wbm_enabled ||
```
isaac-io, line 955: Same here: please use `ASSERT_GT` in an if rather than `ASSERT_TRUE` so that it'll be easier to understand the expected vs. actual in case of failure.

db/memtable_list_test.cc
```diff
@@ -541,313 +541,63 @@ TEST_F(MemTableListTest, GetFromHistoryTest) {
...
-    // Refcount should be 0 after calling TryInstallMemtableFlushResults.
-    // Verify this, by Ref'ing then UnRef'ing:
-    m->Ref();
-    ASSERT_EQ(m, m->Unref());
-    delete m;
-  }
-  to_delete.clear();
+void ValidateWbmUsedCounters(const WriteBufferManager& wb, size_t expected_used,
+                             size_t expected_immutable, size_t expected_freed) {
+  ASSERT_EQ(wb.mutable_memtable_memory_usage(), expected_used);
```
isaac-io, line 548: Shouldn't `expected_used` be named `expected_mutable`?

db/memtable_list.cc
udi-speedb: The original code just set 2 boolean flags without any side effects so the order wasn't at all important.
SetFlushInProgress() / SetFlushCompleted() are now calling alloc tracker's code which uses the state machine and needs the calls in a spceific order.
I don't really understand your point or the reason for it

isaac-io: It doesn't change the behaviour. What I mean is that the original code did it in that order to signify a rollback because `RemoveMemTablesOrRestoreFlags()` is called in `TryInstallMemtableFlushResults()` after `SetFlushCompleted(true)` was called. I'm assuming that you changed the order because of the way the check in `SetFlushCompleted()` relies on `flush_in_progress_` being set to `true` in order to avoid calling abort twice (and the only reason this works is because `flush_is_progress_` is never set to `false` in the good path, when a rollback isn't required). Ideally there wouldn't be two flags, but only a single state machine, but in the current state of things I simply think that the check there should not rely on that but on the value of `flush_completed_` for signifying abort, though I admit that it doesn't change the behaviour, and it was simply a question of regarding the logical ordering of things.

db/memtable.h
```diff
@@ -479,13 +479,27 @@ class MemTable {
...
 
   uint64_t GetFileNumber() const { return file_number_; }
 
   void SetFileNumber(uint64_t file_num) { file_number_ = file_num; }
 
   void SetFlushInProgress(bool in_progress) {
+    if (in_progress && (flush_in_progress_ == false)) {
+      assert(!flush_completed_);
+      mem_tracker_.FreeMemStarted();
+      // In case flush is aborted, notify the memory tracker
```
isaac-io, line 498: Comment in the wrong place. Can you move it inside the `else` branch?

db/memtable.h
```diff
@@ -479,13 +479,27 @@ class MemTable {
...
+    }
+    flush_completed_ = completed;
+  }
 
   uint64_t GetFileNumber() const { return file_number_; }
 
   void SetFileNumber(uint64_t file_num) { file_number_ = file_num; }
 
   void SetFlushInProgress(bool in_progress) {
+    if (in_progress && (flush_in_progress_ == false)) {
```
isaac-io, line 495: Can we simplify the check by wrapping this in `if (in_progress != flush_in_progress_)`, so that inside we'll only need to check if the new value is `true` or not? It took me a couple of reads to understand that this is basically what the checks are all about.

isaac-io (commented):
Sorry for the small bursts of comments. Only 2 minor comments this time, and hopefully it'll be the last round.

db/memtable_list.cc
udi-speedb: So, should I leave it as is or would you like me to change anything?

db/memtable_list.cc
udi-speedb: Same question here

db/memtable_list.cc
isaac-io: While I understand that the order doesn't really matter even if some future programmer fumbles and causes abort to be called twice (since the `AllocTracker` protects against that), I still think that the reordering is unnecessary and creates unneeded confusion for readers of this diff. Additionally, it would still be better IMO if the check in `SetFlushCompleted()` relied on `flush_completed_` rather than `flush_in_progress_` (the current state, where it relies on `flush_in_progress_`, is confusing because `SetFlushCompleted()` sets the former, but not the latter). Also, it would be clearer if `SetFlushInProgress()` made sure that `flush_completed_` is `false` before calling abort.

udi-speedb:
@isaac-io - Please review the last commit. Thanks

isaac-io (approved):
LGTM.

udi-speedb:
I have squashed and pushed the branch

PR #133, opened 2022-08-23
Devops ci/move fuzz to reusable workflow

nathan-bo:


.github/workflows/test_fuzz.yml
```diff
@@ -56,18 +62,18 @@ jobs:
            chmod +x prepfuz.sh
            bash -xv prepfuz.sh
            mkdir -p $GITHUB_WORKSPACE/out/
-           ASAN_OPTIONS=detect_odr_violation=0 $GITHUB_WORKSPACE/fuzz/${{ matrix.name }} 2>&1 | sudo tee $GITHUB_WORKSPACE/out/${{ matrix.name }}.log-$GITHUB_SHA
-           tail -20 $GITHUB_WORKSPACE/out/${{ matrix.name }}.log-$GITHUB_SHA | grep "==AddressSanitizer. Thread limit (4194304 threads) exceeded\. Dying\." || echo "${{ matrix.name }} failed!"
+           ASAN_OPTIONS=detect_odr_violation=0 $GITHUB_WORKSPACE/fuzz/${{ matrix.name }} 2>&1 | sudo tee $GITHUB_WORKSPACE/out/${{ matrix.name }}.log
+           tail -20 $GITHUB_WORKSPACE/out/${{ matrix.name }}.log | grep "==AddressSanitizer. Thread limit (4194304 threads) exceeded\. Dying\." || echo "${{ matrix.name }} failed!"
```
isaac-io, line 66: This will not actually fail the fuzz test. You need a command that doesn't return 0. I suggest adding `&& false` at the end of the line here.

.github/workflows/test_fuzz.yml
```diff
@@ -56,18 +62,18 @@ jobs:
...
            
       - name: Copy ${{ matrix.name }} logs to S3
         run: |
-          aws s3 cp $GITHUB_WORKSPACE/out/${{ matrix.name }}.log-$GITHUB_SHA s3://spdb-github-ci/
+          aws s3 cp $GITHUB_WORKSPACE/out/${{ matrix.name }}.log s3://spdb-github-ci/
```
isaac-io, line 70: How are we going to differentiate between runs if you don't give the file a name that includes the git hash?

.github/workflows/test_fuzz.yml
```diff
@@ -56,18 +62,18 @@ jobs:
            chmod +x prepfuz.sh
            bash -xv prepfuz.sh
```
isaac-io, line 63: Why not simply run the commands instead of outputting them to a file and executing it?

isaac-io (changes requested):

.github/workflows/test_fuzz.yml
nathan-bo: I can add it back no problem, do you want me to?

.github/workflows/test_fuzz.yml
nathan-bo: This is the way it was ran in gitlab ci, I think @maxb-io explained it to me but I can't really remember

.github/workflows/test_fuzz.yml
```diff
@@ -16,15 +21,16 @@ jobs:
...
     steps:
-      - uses: actions/checkout@v3
-
       - name: Pre-build
         run: |
           apt update && apt install -y sudo python3 git clang-tools cmake make automake ucommon-utils libtool gettext pkg-config build-essential clang-10 zlib1g-dev libbz2-dev ninja-build liblzma-dev autoconf libsnappy-dev libzstd-dev liblz4-dev binutils m4 g++-10 unzip
           curl "https://awscli.amazonaws.com/awscli-exe-linux-x86_64.zip" -o "awscliv2.zip"
           unzip awscliv2.zip
           sudo ./aws/install
+          echo "nameserver 8.8.8.8" > /etc/resolv.conf
```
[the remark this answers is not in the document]
nathan-bo: done

.github/workflows/test_fuzz.yml
nathan-bo: fixed, it stayed from tests

.github/workflows/test_fuzz.yml
isaac-io: Please do. There's no need to add it to the file name everywhere, only here when uploading to S3.

.github/workflows/test_fuzz.yml
```diff
@@ -56,18 +62,18 @@ jobs:
            chmod +x prepfuz.sh
            bash -xv prepfuz.sh
            mkdir -p $GITHUB_WORKSPACE/out/
-           ASAN_OPTIONS=detect_odr_violation=0 $GITHUB_WORKSPACE/fuzz/${{ matrix.name }} 2>&1 | sudo tee $GITHUB_WORKSPACE/out/${{ matrix.name }}.log-$GITHUB_SHA
-           tail -20 $GITHUB_WORKSPACE/out/${{ matrix.name }}.log-$GITHUB_SHA | grep "==AddressSanitizer. Thread limit (4194304 threads) exceeded\. Dying\." || echo "${{ matrix.name }} failed!"
+           ASAN_OPTIONS=detect_odr_violation=0 $GITHUB_WORKSPACE/fuzz/${{ matrix.name }} 2>&1 | tee $GITHUB_WORKSPACE/out/${{ matrix.name }}.log
+           tail -20 $GITHUB_WORKSPACE/out/${{ matrix.name }}.log | grep "==AddressSanitizer. Thread limit (4194304 threads) exceeded\. Dying\." || echo "${{ matrix.name }} failed!" && false
```
isaac-io, line 66: Note that due to operator precedence in shell you need to group the last two commands, otherwise the `&&` part would apply regardless of the actual result:

```suggestion
tail -20 $GITHUB_WORKSPACE/out/${{ matrix.name }}.log | grep "==AddressSanitizer. Thread limit (4194304 threads) exceeded\. Dying\." || { echo "${{ matrix.name }} failed!" && false; }
```

nathan-bo: done

.github/workflows/test_fuzz.yml
nathan-bo: done

isaac-io: What you did here wouldn't actually work, because the file isn't created with the hash. What you need is something like the following:

```bash
          aws s3 cp $GITHUB_WORKSPACE/out/${{ matrix.name }}.log s3://spdb-github-ci/$GITHUB_SHA-${{ matrix.name }}.log
```

nathan-bo: changed it, runs properly now

.github/workflows/test_fuzz.yml
```diff
@@ -56,18 +62,18 @@ jobs:
...
-           tail -20 $GITHUB_WORKSPACE/out/${{ matrix.name }}.log-$GITHUB_SHA | grep "==AddressSanitizer. Thread limit (4194304 threads) exceeded\. Dying\." || echo "${{ matrix.name }} failed!"
+           ASAN_OPTIONS=detect_odr_violation=0 $GITHUB_WORKSPACE/fuzz/${{ matrix.name }} 2>&1 | tee $GITHUB_WORKSPACE/out/${{ matrix.name }}.log
+           tail -20 $GITHUB_WORKSPACE/out/${{ matrix.name }}.log | grep "==AddressSanitizer. Thread limit (4194304 threads) exceeded\. Dying\." || { echo "${{ matrix.name }} failed!" && false; }
            
       - name: Copy ${{ matrix.name }} logs to S3
         run: |
-          aws s3 cp $GITHUB_WORKSPACE/out/${{ matrix.name }}.log-$GITHUB_SHA s3://spdb-github-ci/
+          aws s3 cp $GITHUB_WORKSPACE/out/${{ matrix.name }}.log s3://spdb-github-ci/$GITHUB_SHA-${{ matrix.name }}.log
     
   fuzzer_sum:
```
isaac-io, line 72: I'm not sure about the role of this flow. Since the check for fuzzer failure is already done inside the fuzzing workflow, this flow will only be executed on a successful path, which is a waste.

.github/workflows/test_fuzz.yml
```diff
@@ -56,18 +62,18 @@ jobs:
...
            mkdir -p $GITHUB_WORKSPACE/out/
-           ASAN_OPTIONS=detect_odr_violation=0 $GITHUB_WORKSPACE/fuzz/${{ matrix.name }} 2>&1 | sudo tee $GITHUB_WORKSPACE/out/${{ matrix.name }}.log-$GITHUB_SHA
-           tail -20 $GITHUB_WORKSPACE/out/${{ matrix.name }}.log-$GITHUB_SHA | grep "==AddressSanitizer. Thread limit (4194304 threads) exceeded\. Dying\." || echo "${{ matrix.name }} failed!"
+           ASAN_OPTIONS=detect_odr_violation=0 $GITHUB_WORKSPACE/fuzz/${{ matrix.name }} 2>&1 | tee $GITHUB_WORKSPACE/out/${{ matrix.name }}.log
+           tail -20 $GITHUB_WORKSPACE/out/${{ matrix.name }}.log | grep "==AddressSanitizer. Thread limit (4194304 threads) exceeded\. Dying\." || { echo "${{ matrix.name }} failed!" && false; }
            
       - name: Copy ${{ matrix.name }} logs to S3
         run: |
-          aws s3 cp $GITHUB_WORKSPACE/out/${{ matrix.name }}.log-$GITHUB_SHA s3://spdb-github-ci/
+          aws s3 cp $GITHUB_WORKSPACE/out/${{ matrix.name }}.log s3://spdb-github-ci/$GITHUB_SHA-${{ matrix.name }}.log
```
isaac-io, line 70: What's the use of uploading the logs to S3 after the fuzzer has already succeeded? I case of failure this job will not be executed anyway (because the previous job would fail), and in a successful run we have no use for the logs.

nathan-bo: I was told we want to save the logs, is this not correct? I can remove it if there's no use

.github/workflows/test_fuzz.yml
nathan-bo: we could change it if needed

.github/workflows/test_fuzz.yml
isaac-io: The job later removes them, so they are not saved anyway. What we need is a way to store the logs on failure, not on success.

nathan-bo: the clean is only for testing, so that it will automatically clean itself instead of manually doing it. I'll change it to save only when the job fails

.github/workflows/test_fuzz.yml
isaac-io: Let's remove it then. No reason to do redundant work.

.github/workflows/test_fuzz.yml
nathan-bo: made the change, let me know if you want it different

.github/workflows/test_fuzz.yml
```diff
@@ -56,57 +62,10 @@ jobs:
...
-          
-      - name: download db_map_fuzzer.log from s3
-        run: |
-          mkdir -p $GITHUB_WORKSPACE/out
-          aws s3 cp s3://spdb-github-ci/db_map_fuzzer.log $GITHUB_WORKSPACE/out/db_map_fuzzer.log || echo "db_map_fuzzer.log not found"
-          ls -alFh $GITHUB_WORKSPACE/out
-      
-      - name: fuzzersum
+      - name: Copy ${{ matrix.name }} logs to S3 due to failure
+        if: steps.${{ matrix.name }}.outcome != 'success'
```
isaac-io, line 69: This will run on cancellation, which we don't really want. You should probably only run this on failure.

.github/workflows/ci_pipeline.yml
```diff
@@ -104,106 +104,7 @@ jobs:
...
-        run: |
-          echo "~~~~ db_map_fuzzer LOGS ~~~~~"
-          tail -20 $GITHUB_WORKSPACE/out/db_map_fuzzer.log | grep "==AddressSanitizer. Thread limit (4194304 threads) exceeded\. Dying\." || echo "db_map_fuzzer failed!"
-          echo "~~~~ db_fuzzer LOGS ~~~~~"
-          tail -20 $GITHUB_WORKSPACE/out/db_fuzzer.log | grep "==AddressSanitizer. Thread limit (4194304 threads) exceeded\. Dying\." || echo "db_fuzzer failed!"
-          
-      - name: clear s3 bucket
-        run: |
-          aws s3 rm s3://spdb-github-ci/ --recursive --include "*"
+    uses: speedb-io/speedb/.github/workflows/test_fuzz.yml@main
```
isaac-io, line 107: Let's use the [local repository syntax](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#example-using-an-action-in-the-same-repository-as-the-workflow) without pinning to `main`.

isaac-io (changes requested):

.github/workflows/ci_pipeline.yml
nathan-bo: done

.github/workflows/test_fuzz.yml
nathan-bo: changed it to run explicitly if the outcome is 'failure'

isaac-io (approved):

PR #135, opened 2022-08-23
Adding logo to readme

rtpro:


README.md
```diff
@@ -1,3 +1,5 @@
+<img src="speedb-logo.gif" />
+
 # Speedb
```
isaac-io, line 3: This renders way too large on my screens. Can we get someone from marketing to look at it and set the right size? Also, the `Speedb` header seems to be a bit redundant with this logo which already spells it out. I think it should be removed and instead the image should have `Speedb logo` as its alt text (so it's searchable and also visually impaired users as well as those with broken Internet connections can know what it's about).

README.md
```diff
@@ -77,3 +79,5 @@ See the [contributing guide](CONTRIBUTING.md).
 
 ## License
 Speedb is open-source and licensed under the [Apache 2.0 License](LICENSE.Apache).
+
+<img src="speedb-b.gif" />
```
isaac-io, line 83: This seems a bit redundant to me, but as an engineer I'm probably the wrong person to comment on this. Can we get someone from marketing to review this?

maxb-io (approved):

PR #139, opened 2022-08-25
Devops ci/move build to reusable workflow

nathan-bo:


.github/workflows/full-builds.yml
```diff
@@ -0,0 +1,81 @@
...
+      run: |
+        string(TIMESTAMP current_date "%Y-%m-%d-%H;%M;%S" UTC)
+        message("::set-output name=timestamp::${current_date}")
+        
+        
+    - name: ccache cache files
+      uses: actions/cache@v2
+      with:
+        path: ~/.ccache
+        key: ${{runner.os}}-ccache-${{steps.ccache_cache_timestamp.outputs.timestamp}}
```
isaac-io, line 51: Why are we adding a timestamp to the cache key? That makes it so it'll never be hit in later runs. We should use something else (such as the matrix name, as was in the old CI on GitLab).

.github/workflows/full-builds.yml
```diff
@@ -0,0 +1,81 @@
...
+        fi
+        if [ -d "~/.ccache" ]; then
+          echo "Already exists"
+        else
+          mkdir -p ~/.ccache
+          ls ~ | grep cache || echo ""
+          touch ~/.ccache/ccache.txt
+          echo "aaa" > ~/.ccache/ccache.txt
+          ls ~/.ccache
+          cat ~/.ccache/ccache.txt
```
isaac-io, line 69: Why are these needed?

.github/workflows/full-builds.yml
```diff
@@ -0,0 +1,81 @@
...
+        path: ~/.ccache
+        key: ${{runner.os}}-ccache-${{steps.ccache_cache_timestamp.outputs.timestamp}}
+        restore-keys: |
+          ${{runner.os}}-ccache-
+        
+      
+    - name: ${{ matrix.name }}
+      run: |
+        if [ -d "$GITHUB_WORKSPACE/build" ]; then
+          echo >&2 "error: the build directory should not exist" && false NIK
```
isaac-io, line 59: What's with the `NIK` at the end here?

.github/workflows/full-builds.yml
```diff
@@ -0,0 +1,81 @@
...
+          ${{runner.os}}-ccache-
+        
+      
+    - name: ${{ matrix.name }}
+      run: |
+        if [ -d "$GITHUB_WORKSPACE/build" ]; then
+          echo >&2 "error: the build directory should not exist" && false NIK 
+        fi
+        if [ -d "~/.ccache" ]; then
+          echo "Already exists"
```
isaac-io, line 62: If this print is needed, can we make it more meaningful? If it's a leftover from debugging the workflow, let's remove it.

.github/workflows/full-builds.yml
```diff
@@ -0,0 +1,81 @@
...
+            command: CC=clang CXX=clang++ cmake ..  -GNinja
+    container:
+      image: alpine:3.14
+
+    steps:
+    - name: Pre-build
+      run: |
+        env
+        rm -rf /usr/share/dotnet || echo ""
+        df -h
```
isaac-io, line 27: Why are these done? (also note that printing the environment variables is dangerous. It may expose secrets)

.github/workflows/full-builds.yml
```diff
@@ -0,0 +1,81 @@
...
+        apk add git
+        echo "nameserver 8.8.8.8" > /etc/resolv.conf
```
isaac-io, line 29: This should probably be one line up, before `apk add git`.

.github/workflows/full-builds.yml
```diff
@@ -0,0 +1,81 @@
+name: Build
+
+on:
+  workflow_dispatch:
+  workflow_call:
+
```
isaac-io, line 6: Since this workflow doesn't change the repository, can we explicitly set its permissions to `read`?

isaac-io (changes requested):

.github/workflows/full-builds.yml
nathan-bo: In general this is how GitHub  uses cache on their side, you need the key to be unique and the cache hits if a part of the key matches

.github/workflows/full-builds.yml
nathan-bo: This is what we had on gitlab, I'll change it if it's not needed

.github/workflows/full-builds.yml
nathan-bo: by default github actions has read permissions as specified here:
https://docs.github.com/en/actions/security-guides/automatic-token-authentication#permissions-for-the-github_token

.github/workflows/full-builds.yml
nathan-bo: I'll make it more meaningful, no problem

.github/workflows/full-builds.yml
nathan-bo: I'll remove printing of the environment variables

.github/workflows/full-builds.yml
nathan-bo: Will change

.github/workflows/full-builds.yml
isaac-io: That's not what the documentation of the `cache` action says. If no exact match is found for either `key` or `restore-keys`, a new cache is created. See https://docs.github.com/en/actions/using-workflows/caching-dependencies-to-speed-up-workflows#using-the-cache-action

.github/workflows/full-builds.yml
isaac-io: The table you linked to actually says that by default actions have read/write permission with the `GITHUB_TOKEN`.

.github/workflows/full-builds.yml
```diff
@@ -0,0 +1,74 @@
...
+            command: cmake .. -DCMAKE_BUILD_TYPE=Release  -GNinja
+          - mame: clang build
+            command: CC=clang CXX=clang++ cmake ..  -GNinja
+    container:
+      image: alpine:3.14
+
+    steps:
+    - name: Pre-build
+      run: |
+        rm -rf /usr/share/dotnet || echo ""
```
isaac-io, line 25: What's this line for? We don't have dotnet installed in the container anyway.

.github/workflows/full-builds.yml
isaac-io: I was asking specifically about the `NIK` part there. That didn't exist in GitLab.

isaac-io (changes requested):

.github/workflows/full-builds.yml
nathan-bo: Just checked to make sure again, it is taken from Gitlab CI

Screen Shot 2022-08-30 at 16 03 00

nathan-bo: It's left over from one of my tests, removed it

.github/workflows/full-builds.yml
nathan-bo: it shouldn't be there, removed it

.github/workflows/full-builds.yml
nathan-bo: The main problem with not using the timestamps is that you can only use specific cache once for some reason (I can say for sure after testing this as you can see here: https://github.com/speedb-io/speedb/runs/8108257100?check_suite_focus=true
so when you use the timestamp it will get one of the caches that were saved (which doesn't matter since they're all the same) and then lets you use 3 in parallel as you can see here: https://github.com/speedb-io/speedb/actions/runs/2963621327

.github/workflows/full-builds.yml
nathan-bo: after testing, there's no need for any write permissions and I changed it to read-all

isaac-io: Do we really need `read-all`, or can we restrict to `contents: read`?

.github/workflows/full-builds.yml
nathan-bo: I could maybe use a hybrid of the 2, both ${{matrix.name}} and the timestamp so that there could be more than 1 caches available (in case a few jobs run simultaneously they will be able to use the cache) and they will always use the correct cache and not something random

isaac-io: Can you provide an example of where jobs can't use the same cache simultaneously? I'm not sure that I understand why you think such a limitation exists.

nathan-bo: I dont have the old log files but after changing cache v2 to v3 and testing I can't reproduce it. I'll implement the changes and ask for another review

.github/workflows/full-builds.yml
nathan-bo: sure, done

.github/workflows/ci_pipeline.yml
```diff
@@ -8,89 +8,14 @@ on:
...
-        export "CCACHE_DIR=$HOME/.ccache"
-        export "CCACHE_COMPILERCHECK=content"
-        ${{ matrix.command }} -DPORTABLE=1 -DWITH_GFLAGS=1 \
-        -DWITH_ZLIB=1 -DWITH_SNAPPY=1 -DWITH_BZ2=1 -DWITH_LZ4=1 -DWITH_ZSTD=1 \
-        -DWITH_JNI=1 -DJAVA_HOME=/usr/lib/jvm/default-jvm \
-        -DWITH_BENCHMARK_TOOLS=1 -DWITH_CORE_TOOLS=1 -DWITH_TOOLS=1 \
-        -DWITH_TESTS=1 -DWITH_ALL_TESTS=1 -DWITH_EXAMPLES=1
-        ninja
-        
+    uses: speedb-io/speedb/.github/workflows/full-builds.yml@devops-ci/move_build_to_reusable_workflow
```
isaac-io, line 18: This should probably not be pinned to `devops-ci/move_build_to_reusable_workflow`.

nathan-bo: yes, this was only to use a different workflow to test. fixed

isaac-io: Is there a reason not to use the local repository syntax mentioned there? (i.e. `./.github/workflows/full-builds.yml`)

Unless we want to avoid running workflows from the current branch, I see no reason to pin or use the full path.

nathan-bo: we could use it either way, let me know if you want it changed

isaac-io: If there's no reason to pin, let's use the local repository syntax without pinning.

nathan-bo: done

isaac-io (approved):

isaac-io (changes requested):
Note: CI pipeline fails with:

```
Invalid workflow file:
                        .github/workflows/ci_pipeline.yml#L15
The workflow is not valid. 
.github/workflows/ci_pipeline.yml (Line: 15, Col: 3): The workflow must  contain at least one job with no dependencies.
```

https://github.com/speedb-io/speedb/actions/runs/2992824504

nathan-bo:
> Note: CI pipeline fails with:
> 
> ```
> Invalid workflow file:
>                         .github/workflows/ci_pipeline.yml#L15
> The workflow is not valid. 
> .github/workflows/ci_pipeline.yml (Line: 15, Col: 3): The workflow must  contain at least one job with no dependencies.
> ```
> 
> https://github.com/speedb-io/speedb/actions/runs/2992824504

yes, it makes sense since the file which contains the the workflow doesn't exist on main yet since it wasn't pushed

isaac-io:
No, that's not the issue. If you look at the [full CI pipeline file in this PR](https://github.com/speedb-io/speedb/blob/7dfc22019e0a1a0f57d83102b6ebb304977eb7f7/.github/workflows/ci_pipeline.yml) you can immediately see that all of the jobs depend on other jobs (specifically, this PR changes it so that build depends on Sanity, which isn't defined). This prevents a workflow from beginning.

nathan-bo:
> No, that's not the issue. If you look at the [full CI pipeline file in this PR](https://github.com/speedb-io/speedb/blob/7dfc22019e0a1a0f57d83102b6ebb304977eb7f7/.github/workflows/ci_pipeline.yml) you can immediately see that all of the jobs depend on other jobs (specifically, this PR changes it so that build depends on Sanity, which isn't defined). This prevents a workflow from beginning.

Sorry, I was looking at something else.
https://github.com/speedb-io/speedb/pull/84 we will have to approve and push this PR in order for it to be solved (the implementation of sanity to CI)

PR #142, opened 2022-08-28
hash spd bucket get - use key without timestamp

ayulas:


plugin/speedb/memtable/hash_spd_rep.cc
```diff
@@ -703,19 +703,36 @@ class HashLocklessRep : public MemTableRep {
...
-    return MurmurHash(slice.data(), static_cast<int>(slice.size()), 0) %
-           bucket_size_;
+  static size_t GetHash(const Slice& user_key_without_ts) {
+    return MurmurHash(user_key_without_ts.data(),
+                      static_cast<int>(user_key_without_ts.size()), 0);
+  }
+
+  static Slice UserKeyWithoutTimestamp(
+      const Slice internal_key, const MemTableRep::KeyComparator& compare) {
+    auto key_comparator = static_cast<const MemTable::KeyComparator*>(&compare);
```
isaac-io, line 713: Can we use `dynamic_cast` and assert here if `ROCKSDB_USE_RTTI` is set and `NDEBUG` isn't, just to be on the safe side?

```suggestion
#if defined(ROCKSDB_USE_RTTI) && !defined(NDEBUG)
    auto key_comparator = dynamic_cast<const MemTable::KeyComparator*>(&compare);
    assert(key_comparator != nullptr);
#else
    auto key_comparator = static_cast<const MemTable::KeyComparator*>(&compare);
#endif
```

plugin/speedb/memtable/hash_spd_rep.cc
```diff
@@ -703,19 +703,36 @@ class HashLocklessRep : public MemTableRep {
...
-  BucketHeader* GetBucket(const char* key) const {
-    return GetBucket(GetHash(key));
+  BucketHeader* GetBucket(const Slice& internal_key,
+                          const MemTableRep::KeyComparator& comparator) const {
+    const size_t hash =
+        GetHash(UserKeyWithoutTimestamp(internal_key, comparator));
+    BucketHeader* bucket =
+        const_cast<BucketHeader*>(&buckets_[hash % bucket_size_]);
+    return bucket;
   }
```
isaac-io, line 732: Can we inline this in the version that accepts a `char*`, seeing as no API in this file needs to pass an internal key? (the change from memtable key to internal key in line 954 would be unnecessary then)

isaac-io (changes requested):

plugin/speedb/memtable/hash_spd_rep.cc
ayulas: done

plugin/speedb/memtable/hash_spd_rep.cc
ayulas: done

isaac-io (approved):
LGTM.

ayulas:
this. code is the not updated code of hash_spd
this should be ignore and new pull request will be assign to spdb memtable

PR #143, opened 2022-08-30
Add Plugin Tests to builds

mrambacher:
Add a xxx_TESTS variable that contains the tests to build for the plugin.  Create rules in CMake and Make to build the corresponding tests.

Note that the same paradigm can be used to build most/all tests, but was left for a second PR (to not complicate the change)


isaac-io (changes requested):

isaac-io (approved):
LGTM.

Issue #144, opened 2022-08-30
memtable_list: do not roll back memtable flush on CF drop


isaac-io:
When installing memtable flush results, if the flush was unsuccessful (due to an error during the flush or the manifest write) or if the CF was dropped, the memtable is rolled back into a flushable state. However, this is unnecessary since it reactivates the pending flush flag on the relevant CF even though no flush can take place after CF drop, and as in #126 it also causes an issue for the upcoming improvements to the WriteBufferManager immutable memory tracking (#113) because it marks the memtable memory as ready for flush again for a short period before finally getting freed when the memtable is dropped during the destruction of the dropped CF (and this might upset the WBM behaviour because in the future we plan to rely on the amount of memory that isn't already being flushed for triggering flushes rather than on the amount of mutable memory alone as is done today).

This is a continuation of #126 (I simply missed this rollback), and since the rollback only marks the memtable as ready for flush again and clears related flush state, there's no harm in skipping it and just letting the memtable drop in that state.

Yuval-Ariel:
to introduce cf drop, use flag --clear_column_family_one_in . 
note this only works with:
test_batches_snapshots=false
expected_values_dir=None
backup_one_in=0
checkpoint_one_in=0

i would use it with cf_consistency. (blackbox_crash_test_with_atomic_flush)

isaac-io:
Thanks! Can you also post the `db_stress` command that the crash test executed when this failed?

assaf-speedb:
Sure. BTW, the black box failed as well - for the same reason:

Whitebox
[...]

PR #145, opened 2022-08-30
memtable_list: avoid rolling back memtable flush on CF drop (#144)

isaac-io:
This is a continuation of #126 with changes that were missed there in case
a CF drop was encountered after the flush was completed but before or during
manifest write.

Rolling back on CF drop is meaningless because a flush is not possible
in that state, and the only path forwards is for the memtables to be freed
shortly afterwards, so making them available for flush again is useless
at best.

Additionally, this might be needed for the WriteBufferManager changes in
in the future for triggering flushes based on immutable memory as well,
and rolling back flushes causes the memtable memory to become ready for
flush again for a brief period of time until it is dropped, which might
wrongly affect the WBM decisions.

Finally, the code installs a new version even if no changes are made. This
is unnecessary and as such that part is moved into the version-mutating
code path only.

This PR also adds two regression tests, so we could rely on rollback not
being performed on CF drop.

Test Plan: `make check` with the two added tests, as well as running `make crash_test`.

### DIFF
--- a/db/column_family_test.cc
+++ b/db/column_family_test.cc
@@ -1100,6 +1100,189 @@ TEST_P(ColumnFamilyTest, CrashAfterFlush) {
   db_options_.env = env_;
 }
 
+TEST_P(ColumnFamilyTest, DropBeforeInstallResults) {
+  Open();
+  CreateColumnFamilies({"one"});
+
+  // The memtables in the following vector are simply pointers to memtables that
+  // are managed by the CF that is about to be dropped and are collected during
+  // the flush through the sync point callback below. The vector isn't owning
+  // them and access to them is performed only after making sure that they are
+  // still alive (asserting that the amount of immutable memtables that the CF
+  // reports is the same as the amount of memtables that we collected). The
+  // vector is also cleared right after the checks are done in order to avoid
+  // leaking the pointers after they are freed.
+  std::vector<MemTable*> mems;
+
+  SyncPoint::GetInstance()->DisableProcessing();
+  SyncPoint::GetInstance()->ClearAllCallBacks();
+  SyncPoint::GetInstance()->LoadDependency(
+      {{"DBImpl::FlushMemTableToOutputFile:Finish",
+        "ColumnFamilyTest::DropBeforeInstallResults"}});
+  SyncPoint::GetInstance()->SetCallBack(
+      "FlushJob::WriteLevel0Table", [&](void* arg) {
+        auto* memtables = static_cast<autovector<MemTable*>*>(arg);
+        ASSERT_NE(memtables, nullptr);
+        ASSERT_EQ(memtables->size(), 1);
+        for (auto& picked_mem : *memtables) {
+          mems.push_back(picked_mem);
+        }
+        ASSERT_OK(db_->DropColumnFamily(handles_[1]));
+      });
+  SyncPoint::GetInstance()->EnableProcessing();
+
+  ASSERT_OK(Put(1, "foo", "bar"));
+
+  uint64_t num_immutable = 0;
+  ASSERT_TRUE(db_->GetIntProperty(
+      handles_[1], "rocksdb.num-immutable-mem-table", &num_immutable));
+  ASSERT_EQ(num_immutable, 0);
+
+  ASSERT_TRUE(Flush(1).IsColumnFamilyDropped());
+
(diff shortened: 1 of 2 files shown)


db/column_family_test.cc
```diff
@@ -1100,6 +1100,182 @@ TEST_P(ColumnFamilyTest, CrashAfterFlush) {
   db_options_.env = env_;
 }
 
+#ifndef ROCKSDB_LITE  // EventListener is not supported
+TEST_P(ColumnFamilyTest, DropBeforeInstallResults) {
+  struct FlushBeginListener : public EventListener {
```
mrambacher, line 1105: Since EventListener is a class, might be better if this was as well.

db/column_family_test.cc
```diff
@@ -1100,6 +1100,182 @@ TEST_P(ColumnFamilyTest, CrashAfterFlush) {
...
+    void OnFlushBegin(DB* db, const FlushJobInfo& flush_job_info) {
```
mrambacher, line 1106: "override"

db/column_family_test.cc
```diff
@@ -1100,6 +1100,182 @@ TEST_P(ColumnFamilyTest, CrashAfterFlush) {
...
+  SyncPoint::GetInstance()->LoadDependency(
+      {{"DBImpl::FlushMemTableToOutputFile:Finish",
+        "ColumnFamilyTest::DropAfterPickMemtable"}});
+  SyncPoint::GetInstance()->SetCallBack(
+      "FlushJob::WriteLevel0Table", [&](void* arg) {
+        auto* metables = static_cast<autovector<MemTable*>*>(arg);
+        assert(metables);
+        for (auto& picked_mem : *metables) {
+          mems.push_back(picked_mem);
+        }
```
mrambacher, line 1138: Should you guarantee that there is at least one?

And can the variable either me tables or memtables (not metables)?

db/column_family_test.cc
```diff
@@ -1100,6 +1100,182 @@ TEST_P(ColumnFamilyTest, CrashAfterFlush) {
...
+
+  // Ensure the DB closes successfully after this
+  ASSERT_OK(db_->Close());
+  Destroy(descs);
+}
+#endif  // !ROCKSDB_LITE
+
+TEST_P(ColumnFamilyTest, DropAfterPickMemtable) {
+  struct FlushBeginListener : public EventListener {
+    void OnFlushBegin(DB* db, const FlushJobInfo& flush_job_info) {
```
mrambacher, line 1194: class/override, as above

db/column_family_test.cc
isaac-io: Good idea. Thanks for catching the typo!

isaac-io:
I made a copy-pasta mistake and used an EventListener in the wrong test. Fixed it now along with your comments.

db/column_family_test.cc
```diff
@@ -1100,6 +1100,171 @@ TEST_P(ColumnFamilyTest, CrashAfterFlush) {
   db_options_.env = env_;
 }
 
+TEST_P(ColumnFamilyTest, DropBeforeInstallResults) {
+  Open();
+  CreateColumnFamilies({"one"});
+
+  std::vector<MemTable*> mems;
```
mrambacher, line 1107: Since you are storing a pointer to a MemTable, there is no "guarantee" in this code that the stored pointer is not out of scope/deleted later on.  Can you add a comment to the test why this is okay, or maybe store the MemTable autovector instead?

db/column_family_test.cc
```diff
@@ -1100,6 +1100,171 @@ TEST_P(ColumnFamilyTest, CrashAfterFlush) {
...
+  std::shared_ptr<FlushBeginListener> listener =
+      std::make_shared<FlushBeginListener>();
+  db_options_.listeners.push_back(listener);
+
+  Open();
+  CreateColumnFamilies({"one"});
+
+  listener->handle = handles_[1];
+
+  std::vector<MemTable*> mems;
```
mrambacher, line 1199: Same comment/question as above

mrambacher (approved):

db/column_family_test.cc
isaac-io: The autovector is part of the flush job, and gets deleted immediately when the flush job completes, unlike the memtable itself which gets deleted only when its ref-count drops to zero after it is flushed or all of the CF handles dropped. I'll add a comment to explain that.

PR #146, opened 2022-08-30
docs: add a HISTORY file for tracking version changes

isaac-io:
This PR renames the RocksDB HISTORY file to `HISTORY_ROCKSDB.md` and
adds a HISTORY file for tracking Speedb-specific changes.

The PR is made of two commits, the first one for the rename and the other one
for the creation of the file, in order to ease rebasing on future RocksDB versions
later on.


bosmatt:
We would like to have all Speedb and Rocksdb changes in the same History file to preserve the convention and we will have the Release notes under the Releases section in Github.

Issue #156, opened 2022-09-14
build: add a build tag into the version for non-release builds


isaac-io:
For various tasks (especially for the performance tests) we sometimes need to compare builds and tell how they are ordered in the git tree. The current version number is not enough because it's only changed on releases, and while the git hash uniquely identifies a build, it requires access to the git tree in order to find that information, and it becomes useless if the specific commit is later amended or deleted from the tree altogether.

Add an optional build tag to the version and allow setting it during the build (or at configuration time, if using CMake). If it's not set, use a script to calculate a build tag based on the git tree since the last release, falling back to a build tag of `?` in case we aren't able to determine the git tree, which would basically mean that this isn't a release build, and that the git hash should be manually checked as we do today.

isaac-io:
@rtpro @VodOleg please review #157.

erez-speedb:
This tag is now passable and in the collected metadata.

mrambacher:
Is there a better way of doing this?  On my machine this adds 60+ seconds to the make command before we start even start the build.  Is there other information we can collect or a means to collect it faster?

On branches that were committed, the "build date" should have been close to what was required -- it is supposed to capture the date/time of the last git commit.

Also, this PR modifies the "RocksDB version" string when I believe it means to modify the "Speedb" one...

isaac-io:
Is the slowdown happening even in incremental builds, or only after a `make clean`? If the former, we can add caching of the calculated tag so that it's only recalculated if `HEAD` is different than what it was when the tag was calculated. If the latter, I don't have a good solution for that, unfortunately, because we do need to calculate the tag somehow. The idea is not just to capture the date/time, but to allow easily comparing two builds and quickly inferring where they are on the git tree in relation to each other.

As for the version, we actually only have a Speedb version string, since this function is only used for outputting to the LOG file and the tools help text, so we don't really care about backward compatibility, and the format is `<speedb-version> (<rocksdb-version>)` (currently `2.1.0 (7.2.2)`).

mrambacher:
The slowdown happens every time.
time make -j 32 dbg
$DEBUG_LEVEL is 2
Makefile:178: Warning: Compiling in debug mode. Don't use the resulting binary in production
* GEN     make_config.mk
make: Nothing to be done for `dbg'.
make -j 32 dbg  75.01s user 28.32s system 92% cpu 1:51.49 total
markrambacher@marks-mbp speedb % time make -j 32 dbg
$DEBUG_LEVEL is 2
[...]

PR #157, opened 2022-09-14
build: add a version build-tag for non-release builds (#156)

isaac-io:
The build tag can be set during the build (using either the Makefile or the CMake). If it's not provided, and we're not in a release build, it will be calculated using the state of the git tree since the last release tag (for example, for this PR the build tag will be calculated as `(main+17)-(156-build-add-a-build-tag-into-the-version-for-non-release-builds+1)`.

If the git tree state can not be determined, a question mark will be used instead.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1067,6 +1067,24 @@ endif()
 string(REGEX REPLACE "[^0-9a-fA-F]+" "" GIT_SHA "${GIT_SHA}")
 string(REGEX REPLACE "[^0-9: /-]+" "" GIT_DATE "${GIT_DATE}")
 
+option(SPDB_RELEASE_BUILD "Create a release build of Speedb" OFF)
+set(SPDB_BUILD_TAG "" CACHE STRING "Set a specific build tag for this Speedb build")
+
+if(NOT SPDB_RELEASE_BUILD AND "${SPDB_BUILD_TAG}" STREQUAL "")
+  include(FindPython)
+  find_package(Python COMPONENTS Interpreter)
+  if(NOT Python_Interpreter_FOUND)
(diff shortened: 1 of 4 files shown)


build_tools/get_build_tag.py
```diff
@@ -0,0 +1,355 @@
...
+    results = split_nonempty_lines(
+        check_output(
+            ["git", "branch", "--contains", ref, "--format=%(refname:lstrip=2)"]
+        )
+    )
+    for result in results:
+        if result == "main":
+            return (True, result)
+
+        local_candidates.append(result)
```
VodOleg, line 80: The whole git access via shell adds complexity for both invoking operations as well as parsing the results.
Have you considered using a third party abstraction level API like GitPython [https://gitpython.readthedocs.io/en/stable/]

build_tools/get_build_tag.py
```diff
@@ -0,0 +1,355 @@
...
+    for target in local_candidates:
+        boost = -0.5 if hint == (True, target) else 0.0
+        all_candidates.append(
+            (
+                boost
+                + sum(-1.0 for c in local_candidates if is_ancestor_of(c, target)),
+                (True, target),
+            )
+        )
+    all_candidates.sort()
```
VodOleg, line 123: The logic here seem complicated, it is not clear what are you trying to achieve from the comment, perhaps an example along with the desired output may assist understanding this block for future maintanance.

build_tools/get_build_tag.py
isaac-io: Third party libraries need to be installed on user machines, and they don't provide all of the functionality that I'm using here, so we can't just depend on them. Can you point out where specifically you feel that there's unnecessary complexity caused by invoking `git` and parsing its output?

Issue #114, opened 2022-08-04
WriteBufferManager: allow slowing down writes based on total memory usage


isaac-io:
Currently the `WriteBufferManager` (WBM) does not allow slowing down writes in response to memory usage getting close to the prescribed limit. The only mechanism for letting flushes catch up is stopping writes completely by passing true for the `allow_stall` parameter of the WBM constructor. This can lead to oscillation between full write rate and complete stop, which is undesirable as it affects the latency of user writes.

Allow slowing down writes based on memory consumption by having the WBM signal the `WriteController` (WC) (there can be more than one) of the delay requirement. The delay requirement is stored in the WC as part of the Global Delay feature (#346). The same way a CF signals the WC that it has a delay requirement. 

#### To enable this feature:
1. `allow_delays_and_stalls = true`  . in the ctor of WriteBufferManager (previously this flag was named `allow_stalls`)
2. `use_dynamic_delay = true` . 

#### The way the delay requirements are calculated is as follows:
The WBM reports a delay once its memory consumption passes a certain threshold from the quota. That threshold can be controlled by passing `start_delay_percent` to the ctor of the WBM. The default value is 70. Which means that the WBM will start issuing delay requests once the memory consumption of the WBM reaches 70% of its quota. 
The delay is linear throughout the range from threshold to the max quota. The range from start of delay to the quota is divided into 99 steps of delay. (`kMaxDelayedWriteFactor` - 1). E.g. in the 1st step, the delay requirement will be 99/100 * max_delayed_write_rate() and the last step (when the memory almost reached the quota) will result in a delay requirement of 1/100 * max_delayed_write_rate(). max_delayed_write_rate() is the rate the user passed to `delayed_write_rate` (DBOptions) which can also be dynamically changed.

#### note:
The stall logic in the WBM is redundant since the WC already includes logic for stopping writes which can be reused. 
For the first phase, #423 , keep using the stall logic in the WBM (ShouldStall() and WBMStallWrites()) and only add the mechanism for slowing down writes.

udi-speedb:
Pull request: [https://github.com/speedb-io/speedb/pull/164](url)

udi-speedb:
@Guyme The ticket as was actually completed but abandoned.
So, I suggest creating a new ticket once we actually know what we want to do here, and how it fits with other delayed write activities we work on.

Yuval-Ariel:
[...]

PR #164, opened 2022-09-15
Slowdown writes based on WBM's dirty memory usage

udi-speedb:


isaac-io (commented):
I didn't have time to do a thorough review, so I skipped the unit tests and only skimmed the code and left a few comments. I'll go over it properly later and finish the review.

udi-speedb:
Thanks.
Please update once you have completed the entire review. I would like to address all of your comments at once.

udi-speedb:
@isaac-io - Please clearly classify your comments as either _**must**_ (bug, serious design flaw, etc.) and **_may_** in your opinion. Both the comments that you have already written, and those you will add. Thanks

include/rocksdb/write_buffer_manager.h
```diff
@@ -47,9 +60,20 @@ class WriteBufferManager final {
...
+  //
+  // allow_delay: if set to true, it will start delaying of writes when
+  // memory_usage() exceeds the kStartDelayPercentThreshold percent threshold of
+  // the buffer size. The WBM calculates a delay factor that is increasing as
+  // memory_usage() increases. When applicable, the WBM will notify its
+  // registered clients about the applicable delay factor. Clients are expected
+  // to set their respective delayed write rates accordingly. When
+  // memory_usage() reaches buffer_size(), the (optional) WBM stall mechanism
+  // kicks in if enabled. (see allow_stall above)
+  //
```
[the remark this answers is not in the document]
udi-speedb: Obviously I may use the existing allow_stall. However, adding a separate, optional allow_delay argument is preferred IMO:
- The existing WBM stall mechanism remains as is in this commit and allow_stall is the parameter that controls it, so its existing meaning is retained.
- Allows existing users to enable the existing WBM stall mechanism if they wish, without adding this new functionality they may not want for whatever reason.
- Allows us to disable the new behaviour in case we find anything wrong with it (by default it will be disabled). We will hardly have time to understand how this feature behaves in multiple real world scenarios since we have very little time to test it before the release deadline.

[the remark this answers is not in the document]
isaac-io: The problem with adding options is that it's then harder to take them away, and we already have a precedent with the `WriteController` mechanism which doesn't allow this fine-grained control. Also, users who have already opted into stalls likely already care about memory usage more than about the best performance, and at worst we'll cause a degradation in something that's internal code and can be easily reverted, rather than committing something to the public API.

In any case, since all of the arguments are about in-the-wild usage by users, I'm looping in @Guyme and @bosmatt for weighing the options and making a decision.

memtable/write_buffer_manager_test.cc
```diff
@@ -293,6 +295,233 @@ TEST_F(WriteBufferManagerTest, CacheFull) {
 }
 
 #endif  // ROCKSDB_LITE
+
+#define VALIDATE_NOTIFICATION(memory_change_size, expected_state,    \
+                              expected_factor, expect_notification)  \
+  ValidateNotification(__LINE__, memory_change_size, expected_state, \
+                       expected_factor, expect_notification)
+
+class WriteBufferManagerTestWithParms
```
isaac-io, line 304: Nit: typo -> `s/Parms/Params/`

db/db_write_buffer_manager_test.cc
```diff
@@ -94,6 +101,7 @@ TEST_P(DBWriteBufferManagerTest, SharedWriteBufferAcrossCFs2) {
   ASSERT_OK(Put(0, Key(1), DummyString(40000), wo));
   ASSERT_OK(Put(2, Key(1), DummyString(1), wo));
 
+  // std::cout << "Writing 40,000\n";
```
isaac-io, line 104: Nit: please remove debug code here and below.

memtable/write_buffer_manager.cc
```diff
@@ -214,4 +224,102 @@ std::string WriteBufferManager::GetPrintableOptions() const {
...
+      delay_factor =
+          CalcDelayFactor(quota, new_memory_used, usage_start_delay_threshold);
+    } else if (new_usage_state == UsageState::kStop) {
+      delay_factor = kMaxDelayedWriteFactor;
+    }
+    notification_needed = true;
+  } else if (new_usage_state == UsageState::kDelay) {
+    auto memory_used_before = new_memory_used - memory_changed_size;
+    // Calculate & notify only if the change is more than one "step"
+    //
```
isaac-io, line 308: Nit: redundant blank comment line.

isaac-io (commented):
I finished the review. The tests use flushes to test cases where memory is freed and the delay reduced, but I think it needs a test with CF drop as well (in the multi-CF scenario) for validation. Other than that, I left a few very minor comments. Nothing else to add.

include/rocksdb/write_buffer_manager.h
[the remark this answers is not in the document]
udi-speedb: Following a short discussion with @Guyme , @bosmatt, @isaac-io and myself, the decision was to have a single argument that controls both delays (the new mechanism) and stalls (the existing mechanism), **but set its default value to true**.

include/rocksdb/write_buffer_manager.h
```diff
@@ -168,11 +205,19 @@ class WriteBufferManager final {
   // Protects the queue_ and stall_active_.
```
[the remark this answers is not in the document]
udi-speedb: Done

include/rocksdb/write_buffer_manager.h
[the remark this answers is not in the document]
udi-speedb: Done

memtable/write_buffer_manager.cc
```diff
@@ -214,4 +224,102 @@ std::string WriteBufferManager::GetPrintableOptions() const {
...
+}  // namespace
+
+void WriteBufferManager::NotifyUsageIfApplicable(ssize_t memory_changed_size,
+                                                 bool force_notification) {
+  assert(enabled());
+  if (allow_delay_ == false) {
+    return;
+  }
+
+  std::unique_lock<std::mutex> lock(mu_);
```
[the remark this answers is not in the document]
udi-speedb: Done

db/write_controller.h
```diff
@@ -58,14 +67,19 @@ class WriteController {
...
-  void set_delayed_write_rate(uint64_t write_rate) {
+  void set_delayed_write_rate(DelaySource source, uint64_t write_rate) {
     // avoid divide 0
     if (write_rate == 0) {
       write_rate = 1u;
     } else if (write_rate > max_delayed_write_rate()) {
       write_rate = max_delayed_write_rate();
     }
-    delayed_write_rate_ = write_rate;
+    auto source_value = static_cast<uint>(source);
```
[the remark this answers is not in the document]
udi-speedb: Done

db/db_impl/db_impl_open.cc
```diff
@@ -1985,6 +1986,23 @@ Status DBImpl::Open(const DBOptions& db_options, const std::string& dbname,
     }
     handles->clear();
   }
+
+  // TODO - Is this the correct place to register for notifications?
+  // Could the WBM's used memory be > 0?
+  if (s.ok()) {
+    auto wbm = db_options.write_buffer_manager.get();
+    if (wbm && wbm->IsDelayAllowed()) {
```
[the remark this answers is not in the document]
udi-speedb: When delay is disabled, the wbm should not have registered clients, should not notify, etc. 
Cleaner IMO. Leaving as is.

memtable/write_buffer_manager.cc
udi-speedb: Done

db/db_write_buffer_manager_test.cc
udi-speedb: Done

memtable/write_buffer_manager_test.cc
udi-speedb: Done

db/db_impl/db_impl_open.cc
```diff
@@ -1985,6 +1986,23 @@ Status DBImpl::Open(const DBOptions& db_options, const std::string& dbname,
...
+      // Registering regardless of wbm->enabled() since the buffer size may be
+      // set later making the WBM enabled, but we will not re-register again
+      // However, notifications will only be received when the wbm is enabled
+      auto db_impl = static_cast<DBImpl*>(*dbptr);
+      WriteBufferManager::UsageNotificationCb cb =
+          std::bind(&DBImpl::WriteBufferManagerUsageNotification, db_impl,
+                    std::placeholders::_1, std::placeholders::_2);
```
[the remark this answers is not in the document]
udi-speedb: Done

include/rocksdb/write_buffer_manager.h
```diff
@@ -86,6 +112,9 @@ class WriteBufferManager final {
     mutable_limit_.store(new_size * 7 / 8, std::memory_order_relaxed);
     // Check if stall is active and can be ended.
     MaybeEndWriteStall();
+    if (enabled()) {
+      NotifyUsageIfApplicable(0, true /* force notification */);
```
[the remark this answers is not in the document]
udi-speedb: The way the NotifyUsageIfApplicable() is currently written, in case of a change in the delay factor, it will not work (the calculations used to deduce whether we need to recalculate the delay factor will not work).
Changing the buffer size is a rare event (if it occurs at all), so, I find it simpler to unconditionally notify rather than make sure it works for that case as well, without changing the code or writing special code for that case.

memtable/write_buffer_manager.cc
```diff
@@ -214,4 +223,100 @@ std::string WriteBufferManager::GetPrintableOptions() const {
...
+    }
+  }
+  usage_state_ = new_usage_state;
+
+  if (notification_needed) {
+    std::unique_lock<std::mutex> lock(mu_);
+    for (auto& cb_pair : usage_notification_cbs_) {
+      cb_pair.second(new_usage_state, delay_factor);
+    }
+  }
```
isaac-io, line 319: Note that there might be a race here (which was also present before the mutex locking moved, but is now more apparent): If `FreeMem()` and `ReserveMem()` are both competing for a notification, we might reach the following broken state:

1. T1 calls `ReserveMem()`.
2. T1 calculates a Delay state.
3. T2 calls `FreeMem()`.
4. T2 calculates a None state.
5. T2 notifies None state.
6. T1 notifies Delay state.

Now clients are in delay even though the WBM does not require a delay.

memtable/write_buffer_manager.cc
```diff
@@ -274,48 +275,76 @@ void WriteBufferManager::NotifyUsageIfApplicable(ssize_t memory_changed_size,
...
-  }
-  usage_state_ = new_usage_state;
 
-  if (notification_needed) {
-    std::unique_lock<std::mutex> lock(mu_);
-    for (auto& cb_pair : usage_notification_cbs_) {
-      cb_pair.second(new_usage_state, delay_factor);
+    usage_state_ = new_usage_state;
+
+    if UNLIKELY (notification_needed) {
```
isaac-io, line 322: This only works by accident, because the macro expands to something that is enclosed in round brackets. Please use the same style as you used below -- `if (UNLIKELY(...))` -- which is the style used throughout the rest of the code.

memtable/write_buffer_manager.cc
```diff
@@ -274,48 +275,76 @@ void WriteBufferManager::NotifyUsageIfApplicable(ssize_t memory_changed_size,
...
+      // The lock is taken only now to avoid locking on every call (problematic
+      // especially when WBM's ReserveMem() is called, which may be frequent).
+      // Consequently, a race between 2 threads is possible, one freeing
+      // (FreeMem()) and one allocating (ReserveMem()), potentially resulting in
+      // an incorrect notification.
+      // => before actually notifying, make sure the state hasn't changed in a
+      // way that makes
```
isaac-io, line 329: The comment seems broken mid sentence here.

memtable/write_buffer_manager.cc
```diff
@@ -274,48 +275,76 @@ void WriteBufferManager::NotifyUsageIfApplicable(ssize_t memory_changed_size,
...
+      if (lock.owns_lock() == false) {
+        lock.lock();
+
+        // Coarse check (under the lock) to check if a race may have happened.
+        // If so, repeat the evaluation of the new state and need to notify, but
+        // this time under a lock which prevents the race from repeating
+        auto prev_new_memory_used = new_memory_used;
+        new_memory_used = memory_usage();
+        if (UNLIKELY(new_memory_used != prev_new_memory_used)) {
```
isaac-io, line 338: This is actually quite likely to happen in a write heavy workload. It would probably be better to only check if the "step" changed (i.e. `(new_memory_used / change_steps) != (prev_new_memory_used / change_steps)`, perhaps rounded up with something like `(new_memory_used + change_steps - 1) / change_steps` in order to account for fractional results), since that's the only thing that should affect the calculation anyway.

udi-speedb: Do you mean re-evaluation when 
(1) (new_memory_used + change_steps - 1) / change_steps != (prev_new_memory_used / change_steps)
or 
(2) (new_memory_used + change_steps - 1) / change_steps != (prev_new_memory_used + change_steps - 1) / change_steps
And, once the check for re-evaluation if updated, would you say it would be UNLIKELY or not necessarily?

db/db_write_buffer_manager_test.cc
```diff
@@ -7,17 +7,21 @@
 // Use of this source code is governed by a BSD-style license that can be
 // found in the LICENSE file. See the AUTHORS file for names of contributors.
 
+#include <iostream>
```
isaac-io, line 10: This include seems like a leftover.

include/rocksdb/write_buffer_manager.h
```diff
@@ -154,6 +187,24 @@ class WriteBufferManager final {
...
+    return coded_usage_state_.load(std::memory_order_relaxed);
+  }
+  std::pair<UsageState, uint64_t> GetUsageStateInfo() const {
+    return ParseCodedUsageState(GetCodedUsageState());
+  }
+
+  static uint64_t CalcCodedUsageState(UsageState usage_state,
+                                      uint64_t delay_factor);
+  static std::pair<UsageState, uint64_t> ParseCodedUsageState(
+      uint64_t coded_usage_state);
```
isaac-io, line 202: Can we encapsulate this logic in its own type and return that from `GetCodedUsageState()` so that consumers of the API would not need to be aware of the underlying representation?

memtable/write_buffer_manager.cc
```diff
@@ -214,4 +237,143 @@ std::string WriteBufferManager::GetPrintableOptions() const {
...
+
+    if (old_coded_usage_state != new_coded_usage_state) {
+      // Try to update the usage state with the usage state calculated by the
+      // current thread. Failure (has_update_succeeded == false) means one or
+      // more threads have updated the current state, rendering our own
+      // calculation irrelevant. In case has_update_succeeded==false,
+      // old_coded_usage_state will be the value of the state that was updated
+      // by the other thread(s).
+      done = std::atomic_compare_exchange_strong(
+          &coded_usage_state_, &old_coded_usage_state, new_coded_usage_state);
```
isaac-io, line 366: Please use the member function rather than the free function. It makes it easier to see which argument is used for what:

```suggestion
      done = coded_usage_state_.compare_exchange_strong(
          old_coded_usage_state, new_coded_usage_state);
```

isaac-io (commented):
I left some very minor comments, but otherwise LGTM.

isaac-io (approved):

memtable/write_buffer_manager.cc
udi-speedb: Done

db/db_write_buffer_manager_test.cc
udi-speedb: Done

include/rocksdb/write_buffer_manager.h
udi-speedb: Encapsulating it is cleaner, and I have done it, BUT, it has entailed a relatively substantial amount of changes.

isaac-io: Understood. We can defer that for a later PR in that case.

udi-speedb: Too late. The update is "on the fire" and on its way for your next review :-)

isaac-io (approved):
LGTM.

udi-speedb:
@isaac-io - Thanks for the review(s)
I will now squash the commits and push (forced)

udi-speedb:
@isaac-io Please review the updated commit. Thanks

udi-speedb:
@isaac-io - I have pushed the minor fix (forced). Please review. Thanks

db/column_family.cc
```diff
@@ -739,19 +739,26 @@ const double kDelayRecoverSlowdownRatio = 1.4;
 
 namespace {
 // If penalize_stop is true, we further reduce slowdown rate.
+// being one of them. The columen families have their own delay / stop logic
```
isaac-io, line 742: Nit: this comment seems to be missing a line that somehow got pushed down to line 746 below. Also, `columen` -> `column`.

db/write_controller.h
```diff
@@ -21,11 +25,18 @@ class WriteControllerToken;
 // All of the methods here (including WriteControllerToken's destructors) need
 // to be called while holding DB mutex
 class WriteController {
+ public:
+  enum class DelaySource { kCF = 0, kWBM = 1, kNumSources };
+  static constexpr unsigned int DelaySourceValue(DelaySource source) {
```
isaac-io, line 30: Please use `size_t` for the return value, as it's used for indexing into an array.

db/write_controller.h
```diff
@@ -21,11 +25,18 @@ class WriteControllerToken;
...
+    return static_cast<int>(source);
```
isaac-io, line 31: Note: casting to `int`, while return value is `unsigned int`.

db/write_controller.h
```diff
@@ -38,72 +49,131 @@ class WriteController {
...
-      uint64_t delayed_write_rate);
+      DelaySource source, uint64_t delayed_write_rate);
   // When an actor (column family) requests a moderate token, compaction
   // threads will be increased
   std::unique_ptr<WriteControllerToken> GetCompactionPressureToken();
 
   // these three metods are querying the state of the WriteController
   bool IsStopped() const;
-  bool NeedsDelay() const { return total_delayed_.load() > 0; }
+  bool NeedsDelay() const { return (TotalDelayed() > 0); }
```
isaac-io, line 70: Nit: the enclosing round brackets are redundant here and hurt readability.

options/db_options.cc
```diff
@@ -826,13 +826,12 @@ void ImmutableDBOptions::Dump(Logger* log) const {
...
-    ROCKS_LOG_HEADER(log, "                   write_buffer_manager_options:");
-    ROCKS_LOG_HEADER(log, "%s",
-                     write_buffer_manager->GetPrintableOptions().c_str());
-  }
+  ROCKS_LOG_HEADER(
+      log, "                   Options.write_buffer_manager: %p%s%s",
+      write_buffer_manager.get(), (write_buffer_manager.get() ? "\n" : ""),
+      (write_buffer_manager.get()
+           ? write_buffer_manager->GetPrintableOptions().c_str()
+           : ""));
```
isaac-io, line 834: This change doesn't change the functionality, nor make the code more readable. Why is it needed?

isaac-io (approved):
I left some minor comments, but overall LGTM.

udi-speedb:
@isaac-io Thanks for the review.
Have you finished it?
If not, please let me know once you are done so I can fix in a single iteration. 
Thanks

isaac-io:
> @isaac-io Thanks for the review. Have you finished it? If not, please let me know once you are done so I can fix in a single iteration. Thanks

I did. I don't have any more comments, which is why I approved.

udi-speedb:
@Yuval-Ariel This is the PR that was opened for #114. However, this is no longer an issue to be inserted into main as is. Please decide what to do with this PR (probably close it). Thanks

udi-speedb:
@Yuval-Ariel, @Guyme - Is this PR applicable? Shouldn't it be closed in favour of the PR that will include all of @Yuval-Ariel's work, including whatever was used from this PR?

Yuval-Ariel:
yes it will be closed once i open a new one.

Yuval-Ariel:
will be handled by #423

Issue #170, opened 2022-09-21
build: support ccache and sccache in the Makefile


isaac-io:
Using ccache and sccache to speed up build times is already supported by the CMake configuration. It saves a lot of time during the development cycle when debugging or when switching branches and build configurations frequently.

Add the support by detecting the availability of sccache/ccache in the `build_detect_platform` script and using that information in the Makefile.

Guyme:
waiting for @mrambacher review

PR #171, opened 2022-09-21
build: support ccache and sccache in the Makefile build (#170)

isaac-io:
Add support for detecting and using ccache/sccache to the Makefile build in order to speed up build times as it's done in the CMake configuration.

### DIFF
--- a/build_tools/build_detect_platform
+++ b/build_tools/build_detect_platform
@@ -102,6 +102,14 @@ if test -z "$AR"; then
     fi
 fi
 
+if [ "$ROCKSDB_USE_CCACHE" = "1" ]; then
+    if command -v sccache > /dev/null; then
+        CCACHE=sccache
+    elif command -v ccache > /dev/null; then
+        CCACHE=ccache
+    fi
+fi
+
 # Detect OS
 if test -z "$TARGET_OS"; then
     TARGET_OS=`uname -s`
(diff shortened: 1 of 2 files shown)


build_tools/build_detect_platform
```diff
@@ -102,6 +102,12 @@ if test -z "$AR"; then
     fi
 fi
 
+if command -v sccache > /dev/null; then
+    CCACHE=sccache
+elif command -v ccache > /dev/null; then
+    CCACHE=ccache
+fi
+
```
mrambacher, line 110: Please provide a way to turn this on/off.  I have spent too long on issues fighting with the build system to clean up the cache builds.

mrambacher (commented):
There needs to be some switch to enable/disable the use of the cache.  I would prefer if (at least initially) the default is OFF until it can be generally tested and validated by the community.

Does similar work also need to be implemented for CMake or is that already in place?

isaac-io:
> There needs to be some switch to enable/disable the use of the cache. I would prefer if (at least initially) the default is OFF until it can be generally tested and validated by the community.
> 
> Does similar work also need to be implemented for CMake or is that already in place?

I did it this way (without a way to turn it off) because that's how it's implemented on the CMake side. I don't mind adding a switch here, but I'd prefer not to do that for the CMake build because it would change the existing behaviour.

mrambacher (approved):
LGTM!

Issue #172, opened 2022-09-21
Convert the build scripts to POSIX shell


isaac-io:
Some of our CI targets don't have bash by default, and it'd be wasteful to install it when it's not really needed. Bash also might not be available on every user machine, and since there's no real reason to use bash-specific syntax, there's no real reason to depend on bash.

Clean up the scripts used during the build (specifically: `build_detect_platform`, `check_buck_targets.sh`, `check-sources.sh`, and `format-diff.sh`) so they are compatible with POSIX shell and don't require bash.

isaac-io:
Closing due to https://github.com/speedb-io/speedb/pull/173#issuecomment-1291756398

PR #173, opened 2022-09-21
build: convert build-related scripts to POSIX shell (#172)

isaac-io:
Clean up scripts that are used throughout the build and use bash syntax in order to allow running them on platforms that don't have bash installed, since they don't really require any bash-specific features and can easily be converted to POSIX shell.


speedbmike (commented):
I don't like the changes with <<EOF after "if", imho it's better to have them before if. 
Also de morgan instead of if ! makes it less readable. 
Other than that looks ok

tools/rocksdb_dump_test.sh
```diff
@@ -1,9 +1,9 @@
-# shellcheck disable=SC2148
+#!/bin/sh
```
mrambacher, line 1: Why this change?

mrambacher (commented):
I am not sure I understand the purpose of this change.  What platform does not have bash available for it that we want this change?  It also adds another set of changes that will have to be merged and maintained btw Speedb and RocksDB.

tools/rocksdb_dump_test.sh
isaac-io: So that it could be run directly without having to explicitly going through `sh` on the command line.

isaac-io:
> I am not sure I understand the purpose of this change. What platform does not have bash available for it that we want this change? It also adds another set of changes that will have to be merged and maintained btw Speedb and RocksDB.

Alpine Linux, which is used in CI, uses BusyBox by default, not Bash. This is easily solvable there by simply installing Bash, but I wanted to support platforms where that might not be possible. Though I agree that it adds unnecessary maintenance burden, so as long as there's no concrete requirement we probably shouldn't do it. Closing this PR for now.

PR #175, opened 2022-09-21
makefile: rework the dependency graph for faster test runs startup (#…

isaac-io:
…174)

The Makefile has many places where it needlessly invokes a sub-make instead of declaring the dependencies correctly. Rework the dependency graph so that at least running tests doesn't involve invoking make three times.

This also avoids pollution in /dev/shm caused by the needless creation of temporary directories on make restarts and sub-make invocations.


Makefile
```diff
@@ -210,20 +210,34 @@ am__v_AR_1 =
 AM_LINK = $(AM_V_CCLD)$(CXX) -L. $(patsubst lib%.a, -l%, $(patsubst lib%.$(PLATFORM_SHARED_EXT), -l%, $^)) $(EXEC_LDFLAGS) -o $@ $(LDFLAGS) $(COVERAGEFLAGS)
 AM_SHARE = $(AM_V_CCLD) $(CXX) $(PLATFORM_SHARED_LDFLAGS)$@ -L. $(patsubst lib%.$(PLATFORM_SHARED_EXT), -l%, $^) $(EXEC_LDFLAGS) $(LDFLAGS) -o $@
 
+# Only generate make_config.mk during the main make invocation, not on restarts
+# (restarts are caused by Makefiles being updated during the parsing of the Makefile,
+#  which is exactly what happens when make_config.mk is regenerated and included).
+ifeq ($(MAKE_RESTARTS),)
+# Don't regenerate when doing a simple `make clean` invocation and make_config.mk
```
mrambacher, line 217: There are likely other rules that do not require a make_config, such as make watch-log

Makefile
```diff
@@ -949,15 +958,13 @@ else
...
+check_0: all $(parallel_tests)
 	$(AM_V_GEN)printf '%s\n' ''						\
+	  'Running tests in $(TEST_TMPDIR)'		\
 	  'To monitor subtest <duration,pass/fail,name>,'		\
 	  '  run "make watch-log" in a separate window' '';		\
-	{ \
-		printf './%s\n' $(filter-out $(PARALLEL_TEST),$(TESTS)); \
-		find t -name 'run-*' -print; \
-	} \
+		printf './%s\n' $(filter-out $(PARALLEL_TEST),$(TESTS)) $(PARALLEL_TEST:%=t/run-%-*) \
```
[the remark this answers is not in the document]
isaac-io: `find t -name 'run-*'` is replaced by a shell glob expansion with the `$(PARALLEL_TEST:%=t/run-%-*)` part, which turns the `printf` call into something that looks like `printf './%s\n' [list of non parallel tests] t/run-<parallel-test-name>-* ...`.

So if `TESTS` contains `a b c` and `PARALLEL_TESTS` is set to `a b`, we'll get `printf './%s\n' c t/run-a-* t/run-b-*` which the shell will expand to the `run-*` files for the `a` and `b` tests. This way we ensure that we only grab files for what's currently defined in `PARALLEL_TESTS` and not some leftovers that happen to begin with `run-`, and we'll also fail if any of the parallel tests didn't get any `run-*` files generated for it (because the shell glob expansion will fail, so we'll end up passing a non-existing file to GNU Parallel, and it will error out).

Makefile
```diff
@@ -1004,16 +1023,8 @@ dump-log:
...
-	$(AM_V_GEN)if [ "$(J)" != "1" ] && [ "$(PARALLEL_OK)" = "1" ];  \
-	then                                                            \
-	    $(MAKE) T="$$t" check_0;                                    \
-	else                                                            \
-	    for t in $(TESTS); do                                       \
-	      echo "===== Running $$t (`date`)"; ./$$t || exit 1; done;          \
-	fi
-	rm -rf $(TEST_TMPDIR)
+# via the check_0 rule above.  Otherwise, run them sequentially via check_1.
+check: $(if $(shell [ "$(J)" != "1" ] && [ "$(PARALLEL_OK)" = "1" ] && echo 1),check_0,check_1)
```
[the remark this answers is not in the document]
isaac-io: `J` is [set by default to 100%](https://github.com/speedb-io/speedb/blob/4554ea0486b96371bd65c6ec71b7ce9acd5a5962/Makefile#L930).

Makefile
```diff
@@ -210,20 +210,34 @@ am__v_AR_1 =
...
-                  export LDFLAGS="$(EXTRA_LDFLAGS)"; \
-                  export COMPILE_WITH_ASAN="$(COMPILE_WITH_ASAN)"; \
-                  export COMPILE_WITH_TSAN="$(COMPILE_WITH_TSAN)"; \
-                  export COMPILE_WITH_UBSAN="$(COMPILE_WITH_UBSAN)"; \
-                  export PORTABLE="$(PORTABLE)"; \
-                  export ROCKSDB_NO_FBCODE="$(ROCKSDB_NO_FBCODE)"; \
-                  export USE_CLANG="$(USE_CLANG)"; \
-                  export LIB_MODE="$(LIB_MODE)"; \
+$(info * GEN     make_config.mk)
+dummy := $(shell (export ROCKSDB_ROOT="$(CURDIR)" && \
```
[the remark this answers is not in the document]
isaac-io: No real reason. I'll revert.

Makefile
```diff
@@ -1024,9 +1035,9 @@ endif
 endif
 endif
 ifndef SKIP_FORMAT_BUCK_CHECKS
-	$(MAKE) check-format
-	$(MAKE) check-buck-targets
-	$(MAKE) check-sources
+	build_tools/format-diff.sh -c
+	buckifier/check_buck_targets.sh
+	build_tools/check-sources.sh
 endif
```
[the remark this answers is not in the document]
isaac-io: Because then they will execute before the `check` rule, and that means that e.g. formatting issues will prevent us from running the unit tests until they are fixes, which is undesirable during development.

Makefile
isaac-io: I'll try to go over the targets and add such rules to the check.

Makefile
```diff
@@ -210,20 +210,34 @@ am__v_AR_1 =
...
 AM_SHARE = $(AM_V_CCLD) $(CXX) $(PLATFORM_SHARED_LDFLAGS)$@ -L. $(patsubst lib%.$(PLATFORM_SHARED_EXT), -l%, $^) $(EXEC_LDFLAGS) $(LDFLAGS) -o $@
 
+# Only generate make_config.mk during the main make invocation, not on restarts
+# (restarts are caused by Makefiles being updated during the parsing of the Makefile,
+#  which is exactly what happens when make_config.mk is regenerated and included).
+ifeq ($(MAKE_RESTARTS),)
+# Don't regenerate when doing a simple `make clean` invocation and make_config.mk
+# already exists, becuase we will be using the variables from the existing one and
+# deleting it immediately afterwards.
+ifneq ($(strip $(and $(wildcard make_config.mk),$(filter-out clean,$(MAKECMDGOALS) make_config.mk))),make_config.mk)
```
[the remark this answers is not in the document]
isaac-io: I reworded it a bit. Let me know if the new phrasing is better.

Makefile
```diff
@@ -949,15 +961,13 @@ else
 	parallel_redir = >& t/$(test_log_prefix)log-{/} || bash -c "cat t/$(test_log_prefix)log-{/}; exit $$?"
 endif
 
-.PHONY: check_0
-check_0: gen_parallel_tests
+.PHONY: check_0 check_1
+check_0: all $(parallel_tests)
```
mrambacher, line 965: 1. How do the non-parallel tests get executed in this rule?
2. Why does it depend on "all" -- why not TESTS like check_1?

Makefile
```diff
@@ -2310,7 +2312,7 @@ rocksdbjavastaticnexusbundlejar: rocksdbjavageneratepom
 
 # A version of each $(LIBOBJECTS) compiled with -fPIC
 
-jl/%.o: %.cc
+jl/%.o: %.cc make_config.mk
```
mrambacher, line 2315: Just curious if these make more sense as "order-only" rules or if you mean for them to now be dependencies (they were not previously)?

If more things were in this file (like LITE, DEBUG, etc), it might make more sense as a true dependency...

mrambacher (approved):

Makefile
isaac-io: I mean for them to be actual dependencies, because `make_config.mk` contains compilation flags, so if it gets updated (e.g. due to a library being installed or removed; and the changes in `build_detect_platform` ensure that this only happen if there was an actual change in contents) the object files need to be rebuilt. This wasn't the case previously, but I believe it was in error, so I tacked this change on while I was fixing the dependency graph.

Makefile
isaac-io: 1. The test executable is executed once and runs all of the tests serially, instead of a `run-*` script that executes only a single test case from an executable.
2. That's... actually a bug. `check` needs to depend on `all` for running the ldb tests, but I mistakenly moved it to `check_0` instead of making it depend on `$(TESTS)`. Thanks for catching this.

Makefile
mrambacher: I wonder if the %.d rule should also depend on make_config.mk then?

mrambacher (approved):
LGTM.  Thanks!

Makefile
isaac-io: Nice catch! I completely overlooked the `%.d` targets.

PR #176, opened 2022-09-22
154 const size l0 l1 compactions

### LINKED ISSUE #154: const size L0->L1 Compactions
Owner: yuval

problem description : 
During heavy write pressure, L0->L1 compaction in the current implementation may pick many more L0 files than level0_file_num_compaction_trigger. 
Large steps for compaction are one of the major reasons for instability of write performance. The system compacts many L0 files together and moves from a state of large delay to no delay and vice versa.

Furthermore, when the system can not perform L0->L1 compactions it may create an Intra-L0 compaction to reduce the number of files. which could pose an additional problem later on by further increasing the size of L0->L1 compaction.

goal:
To mitigate this issue, the goal of this feature is to limit the size and number of L0 files picked for L0->L1 compaction.

Yuval-Ariel:
closes #154

### DIFF
--- a/db/compact_files_test.cc
+++ b/db/compact_files_test.cc
@@ -118,6 +118,12 @@ TEST_F(CompactFilesTest, L0ConflictsFiles) {
   delete db;
 }
 
+// TODO: add unit tests for:
+// base case - stop adding to L0 files once limit has reached.
+// small files, make sure it keeps adding until it reaches the size limit.
+// make sure - oldest files are picked.
+// test that bulkload still holds.
+
 // to check scenario : 1-3, 5-6, 2-7 L0 files,
 // 1-3 is picked in SetupInitialFiles().
 // will 5-6 be picked in SetupOtherL0FilesIfNeeded.
--- a/db/compact_files_test.cc
+++ b/db/compact_files_test.cc
@@ -118,6 +118,46 @@ TEST_F(CompactFilesTest, L0ConflictsFiles) {
   delete db;
 }
 
+// to check scenario : 1-3, 5-6, 2-7 L0 files,
+// 1-3 is picked in SetupInitialFiles().
+// will 5-6 be picked in SetupOtherL0FilesIfNeeded.
+TEST_F(CompactFilesTest, L0PickExpansion) {
+  Options options;
+  options.create_if_missing = true;
+  options.compaction_style = kCompactionStyleLevel;
+  // Small slowdown and stop trigger for experimental purpose.
+  options.compaction_pri = CompactionPri::kOldestSmallestSeqFirst;
+  options.level0_file_num_compaction_trigger = 4;
+
+  DB* db = nullptr;
+  DestroyDB(db_name_, options);
+  Status s = DB::Open(options, db_name_, &db);
+  assert(s.ok());
+  assert(db);
+  WriteOptions write_options;
+  // create L0 files
+  ASSERT_OK(db->Put(write_options, "foo1", "v1"));
+  ASSERT_OK(db->Put(write_options, "foo3", "v1"));
+  ASSERT_OK(db->Flush(FlushOptions()));
+
+  ASSERT_OK(db->Put(write_options, "foo2", "v1"));
+  ASSERT_OK(db->Put(write_options, "foo5", "v1"));
+  ASSERT_OK(db->Put(write_options, "foo7", "v1"));
+  ASSERT_OK(db->Flush(FlushOptions()));
+
+  ASSERT_OK(db->Put(write_options, "foo5", "v2"));
+  ASSERT_OK(db->Put(write_options, "foo6", "v1"));
+  ASSERT_OK(db->Flush(FlushOptions()));
+
+  ASSERT_OK(db->Put(write_options, "foo8", "v1"));
+  ASSERT_OK(db->Put(write_options, "foo9", "v1"));
+  ASSERT_OK(db->Flush(FlushOptions()));
+
+  ASSERT_OK(static_cast_with_check<DBImpl>(db)->TEST_WaitForBackgroundWork());
+
+  delete db;
+}
+
 TEST_F(CompactFilesTest, MultipleLevel) {
   Options options;
   options.create_if_missing = true;
--- a/db/version_set.cc
+++ b/db/version_set.cc
@@ -1744,10 +1744,7 @@ VersionStorageInfo::VersionStorageInfo(
       finalized_(false),
       force_consistency_checks_(_force_consistency_checks) {
   level0_stop_writes_trigger_ = mutable_cf_options.level0_stop_writes_trigger;
-  level0_size_to_compact_ =
-      mutable_cf_options.write_buffer_size *
-      (mutable_cf_options.level0_stop_writes_trigger / 2 +
-       mutable_cf_options.level0_file_num_compaction_trigger / 2);
+  min_size_to_compact_ = mutable_cf_options.max_bytes_for_level_base;
   // one_compaction_thread = mutable_cf_options
 
   if (ref_vstorage != nullptr) {
@@ -3495,6 +3492,11 @@ void VersionStorageInfo::GetOverlappingInputs(
           }
         }
 
+        // ideal L0->L1 compaction size is the size of L1.
+        // but dont force a compaction size which is below the theoretical L1
+        // size (max_bytes_for_level_base)
+        uint64_t base_level_size = NumLevelBytes(1);
+
         // stop accumulating L0 files for compaction once enough are added.
         // stop when:
         //    1. its not a compaction after bulkload.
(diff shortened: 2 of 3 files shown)


db/version_set.cc
```diff
@@ -3456,7 +3476,22 @@ void VersionStorageInfo::GetOverlappingInputs(
...
+        // stop accumulating L0 files for compaction once enough are added.
+        // 2nd condition is to keep the bulkload behavior of compacting all L0
+        // files at once. and then the 2nd if () is to ensure we only stop
+        // accumulating L0 files if the compaction is big enough.
+        if (inputs->size() == max_num_L0_files_to_compact_ &&
+            level_files_brief_[level].num_files <=
+                level0_stop_writes_trigger_) {
+          if (cur_files_size > min_l0_size_to_compact) {
+            // fprintf(stdout, "Limited to %lu L0 files",
+            //         inputs->size());
```
mrambacher, line 3489: Should this be written to the log?

db/version_set.cc
```diff
@@ -3456,7 +3476,22 @@ void VersionStorageInfo::GetOverlappingInputs(
         iter++;
       } else {
         // if overlap
+
+        // stop accumulating L0 files for compaction once enough are added.
+        // 2nd condition is to keep the bulkload behavior of compacting all L0
+        // files at once. and then the 2nd if () is to ensure we only stop
+        // accumulating L0 files if the compaction is big enough.
+        if (inputs->size() == max_num_L0_files_to_compact_ &&
```
mrambacher, line 3484: Doesn't this need to be >= ?

It seems like inputs could grow beyond max_L0_files based on the file size.  If the files are small, then min_size_to_compact could be < cur_files_size when this condition is encountered by > on the next pass

db/version_set.cc
```diff
@@ -1743,6 +1743,14 @@ VersionStorageInfo::VersionStorageInfo(
       estimated_compaction_needed_bytes_(0),
       finalized_(false),
       force_consistency_checks_(_force_consistency_checks) {
+  max_num_L0_files_to_compact_ =
+      (mutable_cf_options.level0_stop_writes_trigger +
+       mutable_cf_options.level0_file_num_compaction_trigger) /
+      2;
```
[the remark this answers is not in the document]
Yuval-Ariel: thanks

db/version_set.cc
```diff
@@ -1743,6 +1743,14 @@ VersionStorageInfo::VersionStorageInfo(
       estimated_compaction_needed_bytes_(0),
       finalized_(false),
       force_consistency_checks_(_force_consistency_checks) {
+  max_num_L0_files_to_compact_ =
```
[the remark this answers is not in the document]
Yuval-Ariel: sure

db/version_set.cc
Yuval-Ariel: yes i wanted to have some record inserted to the log but there are no immutable options in the scope. what would you suggest?

db/version_set.cc
Yuval-Ariel: right you are. thanks!

db/version_set.cc
Yuval-Ariel: i've decided to leave out such records since we can infer the info from the current L0-L1 compaction info in the LOG

db/version_set.cc
```diff
@@ -3456,7 +3475,20 @@ void VersionStorageInfo::GetOverlappingInputs(
...
+        // 2nd condition is to keep the bulkload behavior of compacting all L0
+        // files at once. and then the 2nd if () is to ensure we only stop
+        // accumulating L0 files if the compaction is big enough.
+        if (inputs->size() >= max_l0_files_to_compact_ &&
+            level_files_brief_[level].num_files <=
+                level0_stop_writes_trigger_) {
+          if (cur_files_size > min_l0_size_to_compact) {
+            return;
+          }
+        }
```
mrambacher, line 3489: Is there a reason this code is not at the end of the "if overlap" case rather than at the beginning?  It seems like you are going to go through more iterations of the loop for no apparent gain.

mrambacher, line 3489: This code looks so that if ALL files in a range overlapped, we may only compact a subset of them.  Is that the intention?  If so, a comment to that affect would be helpful.  If not, this check probably needs moved outside of the iter.end() loop.

db/version_set.cc
```diff
@@ -3456,7 +3475,20 @@ void VersionStorageInfo::GetOverlappingInputs(
...
+
+        // stop accumulating L0 files for compaction once enough are added.
+        // 2nd condition is to keep the bulkload behavior of compacting all L0
+        // files at once. and then the 2nd if () is to ensure we only stop
+        // accumulating L0 files if the compaction is big enough.
+        if (inputs->size() >= max_l0_files_to_compact_ &&
+            level_files_brief_[level].num_files <=
+                level0_stop_writes_trigger_) {
+          if (cur_files_size > min_l0_size_to_compact) {
+            return;
```
mrambacher, line 3487: For testing purposes, I believe you are going to want to add a SyncPoint here so that you can see from the testing perspective that the code stopped at the right size/number of files.

db/compact_files_test.cc
```diff
@@ -118,6 +118,46 @@ TEST_F(CompactFilesTest, L0ConflictsFiles) {
   delete db;
 }
 
+// to check scenario : 1-3, 5-6, 2-7 L0 files,
+// 1-3 is picked in SetupInitialFiles().
+// will 5-6 be picked in SetupOtherL0FilesIfNeeded.
```
mrambacher, line 123: I do not see how this test is confirming the check scenario.  I think you most likely are going to need some sort of a sync point?

db/compact_files_test.cc
```diff
@@ -118,6 +118,46 @@ TEST_F(CompactFilesTest, L0ConflictsFiles) {
...
+TEST_F(CompactFilesTest, L0PickExpansion) {
```
mrambacher, line 124: Should there be a comparable test with FIFO compaction?

db/compact_files_test.cc
```diff
@@ -118,6 +118,46 @@ TEST_F(CompactFilesTest, L0ConflictsFiles) {
...
+  options.create_if_missing = true;
+  options.compaction_style = kCompactionStyleLevel;
+  // Small slowdown and stop trigger for experimental purpose.
+  options.compaction_pri = CompactionPri::kOldestSmallestSeqFirst;
+  options.level0_file_num_compaction_trigger = 4;
+
+  DB* db = nullptr;
+  DestroyDB(db_name_, options);
+  Status s = DB::Open(options, db_name_, &db);
+  assert(s.ok());
```
mrambacher, line 135: ASSERT_OK(s)

db/compact_files_test.cc
```diff
@@ -118,6 +118,46 @@ TEST_F(CompactFilesTest, L0ConflictsFiles) {
   delete db;
 }
 
+// to check scenario : 1-3, 5-6, 2-7 L0 files,
+// 1-3 is picked in SetupInitialFiles().
+// will 5-6 be picked in SetupOtherL0FilesIfNeeded.
+TEST_F(CompactFilesTest, L0PickExpansion) {
+  Options options;
+  options.create_if_missing = true;
+  options.compaction_style = kCompactionStyleLevel;
```
mrambacher, line 127: You should explicitly set the parameters you are relying on, and maybe to values other than the default.

db/version_set.cc
```diff
@@ -3438,6 +3446,17 @@ void VersionStorageInfo::GetOverlappingInputs(
...
+
+  // sortedness is required for https://github.com/speedb-io/speedb/issues/154.
+  // order of files is newest to oldest.
+  // files seq: 21 - 30 , 11 - 20 , 1 - 10
+  for (size_t i = 1; i < level_files_brief_[level].num_files; i++) {
+    assert(files_[0][i - 1]->fd.smallest_seqno >=
+           files_[0][i]->fd.largest_seqno);
+  }
+  // to make sure that files are picked from oldest to newest.
+  index.reverse();
```
mrambacher, line 3459: Rather than the reverse here, why not put the files into index in the opposite order above (line 3446)?

db/version_set.cc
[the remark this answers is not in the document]
Yuval-Ariel: @mrambacher , is this resolved?

db/version_set.cc
Yuval-Ariel: my intention was to make the code added into rocksdb as external as possible.

db/version_set.cc
Yuval-Ariel: sure, it seemed more logical to me to do the check before inserting a new item but moving it to the end does save an iteration.

Yuval-Ariel: sure i'll add a comment

db/version_set.cc
Yuval-Ariel: done

Issue #177, opened 2022-09-23
Parallel make check fails on the first run


isaac-io:
After the changes in #141, when running `make check` for the first time after a `make clean` with GNU Parallel installed, the following message is printed by GNU Parallel and the run fails:

```
Option tmpdir requires an argument
```

Subsequent invocations of `make check` succeed.

### To Reproduce
1. `make clean` 
2. Ensure GNU Parallel is installed
3. `make -j$(nproc) check`

### Expected behavior
Unit tests successfully run on the first `make check` invocation after a `make clean`

### System (please complete the following information):
 - OS: Ubuntu 22.04
 - Compiler: GCC 12.2.0

### Additional context
N/A.

PR #178, opened 2022-09-23
makefile: fix parallel `make check` failing on the first run (#177)

isaac-io:
After the changes in #141, parallel `make check` started failing with the following GNU Parallel error:

> Option tmpdir requires an argument

This is caused by the fact that `test_config.mk` uses `?=` for setting `TEST_TMPDIR` and on make restarts this doesn't work for some reason (perhaps having to do with empty `TEST_TMPDIR` being somehow exported).

Replace it with a check for empty value and unconditional assignment instead. While at it, include `test_config.mk` only on make restarts for check targets instead of forcing everything that makes use of `TEST_TMPDIR` to reuse the unit tests configuration.

Also use 6 Xs for the mktemp template in order to support BusyBox mktemp and maybe some others as well, that require the template to end with 6 Xs exactly.


common.mk
```diff
@@ -27,11 +28,11 @@ ifeq ($(BASE_TMPDIR),)
 BASE_TMPDIR :=/tmp
 endif
 # Use /dev/shm on Linux if it has the sticky bit set (otherwise, /tmp or other
-# base dir), and create a randomly-named rocksdb.XXXX directory therein.
+# base dir), and create a randomly-named rocksdb.XXXXXX directory therein.
```
mrambacher, line 31: Is there a reason why you changed the randomness/added more XXX?  A comment somewhere (either in the source or the PR) would suffice.

mrambacher (approved):
LGTM.

common.mk
isaac-io: I'll add a comment in the Makefile, but I did explain it in the commit message:

>Also use 6 Xs for the mktemp template in order to support BusyBox mktemp and maybe some others as well, that require the template to end with 6 Xs exactly.

PR #185, opened 2022-09-29
155 proactive wbm flushes

udi-speedb:
@isaac-io - This is a preliminary version not yet ready.
However, due to the time pressure, I would really appreciate your review.
Please focus on the production code parts and leave the unit testing part for later.
Thanks

### DIFF
--- a/db/db_impl/db_impl_compaction_flush.cc
+++ b/db/db_impl/db_impl_compaction_flush.cc
@@ -8,6 +8,7 @@
 // found in the LICENSE file. See the AUTHORS file for names of contributors.
 #include <cinttypes>
 #include <deque>
+#include <limits>
 
 #include "db/builder.h"
 #include "db/db_impl/db_impl.h"
@@ -2827,12 +2828,17 @@ Status DBImpl::BackgroundFlush(bool* made_progress, JobContext* job_context,
           bg_job_limits.max_compactions, bg_flush_scheduled_,
           bg_compaction_scheduled_);
     }
+    *reason = bg_flush_args[0].cfd_->GetFlushReason();
+    if (write_buffer_manager_) {
+      write_buffer_manager_->FlushStarted(
+          *reason == FlushReason::kWriteBufferManagerInitiated);
+    }
+
     status = FlushMemTablesToOutputFiles(bg_flush_args, made_progress,
                                          job_context, log_buffer, thread_pri);
     TEST_SYNC_POINT("DBImpl::BackgroundFlush:BeforeFlush");
     // All the CFDs in the FlushReq must have the same flush reason, so just
     // grab the first one
-    *reason = bg_flush_args[0].cfd_->GetFlushReason();
     for (auto& arg : bg_flush_args) {
       ColumnFamilyData* cfd = arg.cfd_;
       if (cfd->UnrefAndTryDelete()) {
@@ -2917,6 +2923,10 @@ void DBImpl::BackgroundCallFlush(Env::Priority thread_pri) {
     assert(num_running_flushes_ > 0);
     num_running_flushes_--;
     bg_flush_scheduled_--;
+    if (write_buffer_manager_) {
+      write_buffer_manager_->FlushEnded(
+          reason == FlushReason::kWriteBufferManagerInitiated);
+    }
     // See if there's more work to be done
     MaybeScheduleFlushOrCompaction();
     atomic_flush_install_cv_.SignalAll();
@@ -3841,6 +3851,178 @@ Status DBImpl::WaitForCompact(bool wait_unscheduled) {
     bg_cv_.Wait();
   }
   return error_handler_.GetBGError();
+}
+
+bool DBImpl::InitiateMemoryManagerFlushRequest(size_t min_size_to_flush) {
+  if (shutdown_initiated_) {
+    return false;
+  }
+
+  FlushOptions flush_options;
+  flush_options.allow_write_stall = true;
+  flush_options.wait = false;
+
+  if (immutable_db_options_.atomic_flush) {
+    return InitiateMemoryManagerFlushRequestAtomicFlush(min_size_to_flush,
+                                                        flush_options);
+  } else {
+    return InitiateMemoryManagerFlushRequestNonAtomicFlush(min_size_to_flush,
+                                                           flush_options);
+  }
+}
+
+bool DBImpl::InitiateMemoryManagerFlushRequestAtomicFlush(
+    size_t min_size_to_flush, const FlushOptions& flush_options) {
+  assert(immutable_db_options_.atomic_flush);
+
+  autovector<ColumnFamilyData*> cfds;
+  {
+    InstrumentedMutexLock lock(&mutex_);
+
+    SelectColumnFamiliesForAtomicFlush(&cfds);
+    if (cfds.empty()) {
+      return false;
+    }
+
+    // min_size_to_flush may be 0.
+    // Since proactive flushes are active only once recovery is complete =>
+    // SelectColumnFamiliesForAtomicFlush() will keep cf-s in cfds collection
+    // only if they have a non-empty mutable memtable or any immutable memtable
+    // => skip the checks and just flush the selected cf-s.
+    if (min_size_to_flush > 0) {
+      size_t total_size_to_flush = 0U;
+      for (const auto& cfd : cfds) {
+        // Once at least one CF has immutable memtables, we will flush
+        if (cfd->imm()->NumNotFlushed() > 0) {
+          // Guarantee a atomic flush will occur
+          total_size_to_flush = min_size_to_flush;
+          break;
+        } else if (cfd->mem()->IsEmpty() == false) {
+          total_size_to_flush += cfd->mem()->ApproximateMemoryUsage();
+        }
+      }
+      if (total_size_to_flush < min_size_to_flush) {
+        return false;
+      }
+    }
+  }
+
+  ROCKS_LOG_INFO(immutable_db_options_.info_log,
+                 "write buffer manager initiated Atomic flush started current "
+                 "usage %lu out of %lu",
+                 cfds.front()->write_buffer_mgr()->memory_usage(),
+                 cfds.front()->write_buffer_mgr()->buffer_size());
+
+  TEST_SYNC_POINT(
+      "DBImpl::InitiateMemoryManagerFlushRequestAtomicFlush::BeforeFlush");
+  auto s = AtomicFlushMemTables(cfds, flush_options,
+                                FlushReason::kWriteBufferManagerInitiated);
+
+  ROCKS_LOG_INFO(
+      immutable_db_options_.info_log,
+      "write buffer manager initiated Atomic flush finished, status: %s",
+      s.ToString().c_str());
+  return s.ok();
+}
+
+bool DBImpl::InitiateMemoryManagerFlushRequestNonAtomicFlush(
+    size_t min_size_to_flush, const FlushOptions& flush_options) {
+  assert(immutable_db_options_.atomic_flush == false);
+
+  // Pick the "oldest" CF that meets one of the following:
+  // 1. Has at least one IMMUTABLE memtable (=> already has a memtable that
+  //     should be flushed); Or
+  // 2. Has a MUTABLE memtable > min size to flush
+  //
+  // However, care must be taken to avoid starving a CF which has data to flush
+  // (=> and associated WAL) but, to which there is not much writing. So, in
+  // case we find such a CF that is lagging enough in the number of flushes it
+  // has undergone, relative to the cf picked originally, we will pick it
+  // instead, regardless of its mutable memtable size.
+
+  // The CF picked based on min min_size_to_flush
+  ColumnFamilyData* orig_cfd_to_flush = nullptr;
+  // The cf to actually flush (possibly == orig_cfd_to_flush)
+  ColumnFamilyData* cfd_to_flush = nullptr;
+  SequenceNumber seq_num_for_cf_picked = kMaxSequenceNumber;
+
+  {
+    InstrumentedMutexLock lock(&mutex_);
+
+    // First pick the oldest CF with data to flush that meets
+    // the min_size_to_flush condition
+    for (auto* cfd : *versions_->GetColumnFamilySet()) {
+      if (cfd->IsDropped()) {
+        continue;
+      }
+      if ((cfd->imm()->NumNotFlushed() != 0) ||
+          ((cfd->mem()->IsEmpty() == false) &&
+           (cfd->mem()->ApproximateMemoryUsage() >= min_size_to_flush))) {
+        uint64_t seq = cfd->mem()->GetCreationSeq();
+        if (cfd_to_flush == nullptr || seq < seq_num_for_cf_picked) {
+          cfd_to_flush = cfd;
+          seq_num_for_cf_picked = seq;
+        }
+      }
+    }
+
+    if (cfd_to_flush == nullptr) {
+      return false;
+    }
+
+    orig_cfd_to_flush = cfd_to_flush;
+
+    // A CF was picked. Now see if it should be replaced with a lagging CF
+    for (auto* cfd : *versions_->GetColumnFamilySet()) {
+      if (cfd == orig_cfd_to_flush) {
+        continue;
+      }
+
+      if ((cfd->imm()->NumNotFlushed() != 0) ||
+          (cfd->mem()->IsEmpty() == false)) {
+        // The first lagging CF is picked. There may be another lagging CF that
+        // is older, however, that will be fixed the next time we evaluate.
+        if (cfd->GetNumQueuedForFlush() +
+                ColumnFamilyData::kLaggingFlushesThreshold <
+            orig_cfd_to_flush->GetNumQueuedForFlush()) {
+          // Fix its counter so it is considered lagging again only when
+          // it is indeed lagging behind
+          cfd->SetNumTimedQueuedForFlush(
+              orig_cfd_to_flush->GetNumQueuedForFlush() - 1);
+          cfd_to_flush = cfd;
+          break;
+        }
+      }
+    }
+
+    autovector<ColumnFamilyData*> cfds{cfd_to_flush};
+    MaybeFlushStatsCF(&cfds);
+  }
+
+  ROCKS_LOG_INFO(immutable_db_options_.info_log,
+                 "[%s] write buffer manager initiated flush "
+                 "started current "
+                 "usage %lu out of %lu, min-size:%lu, seq:%lu, "
+                 "num-flushes:%lu, orig-cf:%s num-flushes:%lu",
+                 cfd_to_flush->GetName().c_str(),
+                 cfd_to_flush->write_buffer_mgr()->memory_usage(),
+                 cfd_to_flush->write_buffer_mgr()->buffer_size(),
+                 min_size_to_flush, seq_num_for_cf_picked,
+                 cfd_to_flush->GetNumQueuedForFlush(),
+                 orig_cfd_to_flush->GetName().c_str(),
+                 orig_cfd_to_flush->GetNumQueuedForFlush());
+
+  TEST_SYNC_POINT(
+      "DBImpl::InitiateMemoryManagerFlushRequestNonAtomicFlush::BeforeFlush");
+  auto s = FlushMemTable(cfd_to_flush, flush_options,
+                         FlushReason::kWriteBufferManagerInitiated);
+
+  ROCKS_LOG_INFO(
+      immutable_db_options_.info_log,
+      "[%s] write buffer manager initialize flush finished, status: %s\n",
+      cfd_to_flush->GetName().c_str(), s.ToString().c_str());
+
+  return s.ok();
 }
 
 }  // namespace ROCKSDB_NAMESPACE
--- a/include/rocksdb/listener.h
+++ b/include/rocksdb/listener.h
@@ -169,6 +169,7 @@ enum class FlushReason : int {
   // will not be called to avoid many small immutable memtables.
   kErrorRecoveryRetryFlush = 0xc,
   kWalFull = 0xd,
+  kWriteBufferManagerInitiated = 0xe,
 };
 
 // TODO: In the future, BackgroundErrorReason will only be used to indicate
--- a/include/rocksdb/write_buffer_manager.h
+++ b/include/rocksdb/write_buffer_manager.h
@@ -16,12 +16,20 @@
 #include <condition_variable>
 #include <cstddef>
 #include <list>
+#include <memory>
 #include <mutex>
+#include <thread>
+#include <unordered_map>
+#include <utility>
+#include <vector>
 
 #include "rocksdb/cache.h"
 
 namespace ROCKSDB_NAMESPACE {
+struct Options;
 class CacheReservationManager;
+class InstrumentedMutex;
+class InstrumentedCondVar;
 
 // Interface to block and signal DB instances, intended for RocksDB
 // internal use only. Each DB instance contains ptr to StallInterface.
@@ -36,6 +44,22 @@ class StallInterface {
 
 class WriteBufferManager final {
  public:
+  // TODO: Need to find an alternative name as it is misleading
+  // we start flushes in kStartFlushPercentThreshold / number of parallel
+  // flushes
+  static constexpr uint64_t kStartFlushPercentThreshold = 80U;
+
+  struct FlushInitiationOptions {
+    static constexpr size_t kDfltMaxNumParallelFlushes = 4U;
+
+    FlushInitiationOptions() {}
+    size_t max_num_parallel_flushes = kDfltMaxNumParallelFlushes;
+  };
+
+  static constexpr bool kDfltAllowStall = false;
+  static constexpr bool kDfltInitiateFlushes = true;
+
+ public:
   // Parameters:
   // _buffer_size: _buffer_size = 0 indicates no limit. Memory won't be capped.
   // memory_usage() won't be valid and ShouldFlush() will always return true.
@@ -47,9 +71,20 @@ class WriteBufferManager final {
   // allow_stall: if set true, it will enable stalling of writes when
   // memory_usage() exceeds buffer_size. It will wait for flush to complete and
   // memory usage to drop down.
-  explicit WriteBufferManager(size_t _buffer_size,
-                              std::shared_ptr<Cache> cache = {},
-                              bool allow_stall = false);
+  //
+  // initiate_flushes: if set true, the WBM will proactively request registered
+  // DB-s to flush. The mechanism is based on initiating an increasing number of
+  // flushes as the memory usage increases. If set false, WBM clients need to
+  // call ShouldFlush() and the WBM will indicate if current memory usage merits
+  // a flush. Currently the ShouldFlush() mechanism is used only in the
+  // write-path of a DB.
+  explicit WriteBufferManager(
+      size_t _buffer_size, std::shared_ptr<Cache> cache = {},
+      bool allow_stall = kDfltAllowStall,
+      bool initiate_flushes = kDfltInitiateFlushes,
+      const FlushInitiationOptions& flush_initiation_options =
+          FlushInitiationOptions());
+
   // No copying allowed
   WriteBufferManager(const WriteBufferManager&) = delete;
   WriteBufferManager& operator=(const WriteBufferManager&) = delete;
@@ -108,13 +143,18 @@ class WriteBufferManager final {
 
     // Check if stall is active and can be ended.
     MaybeEndWriteStall();
+    if (enabled()) {
+      if (initiate_flushes_) {
+        InitFlushInitiationVars(new_size);
+      }
+    }
   }
 
   // Below functions should be called by RocksDB internally.
 
   // Should only be called from write thread
   bool ShouldFlush() const {
-    if (enabled()) {
+    if ((initiate_flushes_ == false) && enabled()) {
       if (mutable_memtable_memory_usage() >
           mutable_limit_.load(std::memory_order_relaxed)) {
         return true;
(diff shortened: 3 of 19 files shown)


memtable/write_buffer_manager.cc
```diff
@@ -243,7 +291,382 @@ std::string WriteBufferManager::GetPrintableOptions() const {
...
+    // Two iterations:
+    // 1. Flushes of a min size.
+    // 2. Flushes of any size
+    constexpr size_t kNumIters = 2U;
+    const std::array<size_t, kNumIters> kMinFlushSizes{min_flush_size_, 0U};
+
+    auto iter = 0U;
+    while ((iter < kMinFlushSizes.size()) && (num_flushes_to_initiate_ > 0U)) {
+      auto num_repeated_failures_to_initiate = 0U;
+      while ((num_repeated_failures_to_initiate < num_initiators_) &&
```
isaac-io, line 537: This is the only place that `num_initiators_` is read, and it's done under the mutex, so I'm not sure why it's needed (along with the redundancy and bookkeeping that it entails) when `flush_initiators_.size()` can be used directly.

memtable/write_buffer_manager.cc
```diff
@@ -243,7 +291,382 @@ std::string WriteBufferManager::GetPrintableOptions() const {
...
+uint64_t WriteBufferManager::FindInitiator(void* initiator) const {
+  // Assumes lock is held on the flushes_mu_
+
+  auto initiator_idx = kInvalidInitiatorIdx;
+  for (auto i = 0U; i < flush_initiators_.size(); ++i) {
+    if (flush_initiators_[i].initiator == initiator) {
+      initiator_idx = i;
+      break;
+    }
+  }
```
isaac-io, line 644: Why are we using a vector and iterating over it in order to find things instead of using an `unordered_map` with the initiator as the key?

memtable/write_buffer_manager.cc
```diff
@@ -243,7 +291,382 @@ std::string WriteBufferManager::GetPrintableOptions() const {
...
+      options.max_write_buffer_number / 2;
+  int compaction_trigger = options.level0_file_num_compaction_trigger;
+  if (compaction_trigger == 0) {
+    compaction_trigger = options.level0_file_num_compaction_trigger = 4;
+  }
+  // URQ - Why and should it also be part of the options sanitization?
+  // // options.max_background_compactions = options.max_background_flushes * 2;
+  // // options.max_background_jobs =
+  // //   options.max_background_flushes + options.max_background_compactions;
+}
```
isaac-io, line 670: Why is this defined here, rather than as an `Options` method, similar to e.g. `OptimizeForPointLookup()`?

memtable/write_buffer_manager.cc
```diff
@@ -243,7 +291,382 @@ std::string WriteBufferManager::GetPrintableOptions() const {
...
+
+void WriteBufferManager::FlushEnded(bool /* wbm_initiated */) {
+  if (!enabled()) {
+    return;
+  }
+
+  std::unique_lock<std::mutex> lock(flushes_mu_);
+
+  assert(num_running_flushes_ > 0U);
```
isaac-io, line 607: Note that the WBM can be enabled after a flush has been initiated but before it has finished, causing this assertion to fire.

memtable/write_buffer_manager.cc
```diff
@@ -243,7 +291,382 @@ std::string WriteBufferManager::GetPrintableOptions() const {
...
+      while ((num_repeated_failures_to_initiate < num_initiators_) &&
+             (num_flushes_to_initiate_ > 0U)) {
+        bool was_flush_initiated = false;
+        // Unlocking the flushed_mu_ since flushing (via the initiator cb) may
+        // call a WBM service (e.g., ReserveMem()), that, in turn, needs to lock
+        // the same mutex => will get stuck
+        lock.unlock();
+
+        {
+          std::unique_lock<std::mutex> initiators_lock(flushes_initiators_mu_);
```
isaac-io, line 546: Why do we need the separate lock here? This is only used for excluding registration and deregistration, and since they happen on DB open and close, there's no reason not to wait until the flush thread finished a cycle.

The additional lock here may also unnecessarily put the flush initiator thread to sleep. I think it's better to delay opening a closing of the DB than to hinder the normal operation of already opened DBs.

memtable/write_buffer_manager.cc
```diff
@@ -243,7 +291,382 @@ std::string WriteBufferManager::GetPrintableOptions() const {
...
+          auto& initiator = flush_initiators_[next_candidate_initiator_idx_];
+          next_candidate_initiator_idx_ = CalcNextCandidateInitiatorIdx();
+          was_flush_initiated = initiator.cb(kMinFlushSizes[iter]);
```
isaac-io, line 549: Why do we request a minimum size here? If we're trying to choose the "best" candidate (in terms of freeing the most memory), wouldn't it be easier to simply poll the DBs for their resident memory size and choose the one with the most potential to free? If round robin behaviour is really needed we can also take that into account by keeping a score, but otherwise we'll be blindly initiating flushes, potentially causing performance issues by initiating many small flushes rather than allowing more breathing room.

db/db_impl/db_impl_compaction_flush.cc
```diff
@@ -3786,4 +3795,100 @@ Status DBImpl::WaitForCompact(bool wait_unscheduled) {
...
+        if ((cfd->imm()->NumNotFlushed() != 0) ||
+            (!cfd->mem()->IsEmpty() &&
+             (cfd->mem()->ApproximateMemoryUsage() >= min_size_to_flush))) {
+          // We only consider active mem table, hoping immutable memtable is
+          // already in the process of flushing.
+          uint64_t seq = cfd->mem()->GetCreationSeq();
+          if (cfd_picked == nullptr || seq < seq_num_for_cf_picked) {
+            cfd_picked = cfd;
+            seq_num_for_cf_picked = seq;
+          }
```
isaac-io, line 3840: Why are we choosing CFs by age? The only benefit to doing so that I can think of is that we'll be unlikely to choose the same CF again next time around (though it's not guaranteed, because the WBM is not the only initiator of memtable switch, which is what we're relying on here), but other than that we might be causing many small flushes instead of giving ourselves more breathing room.

include/rocksdb/write_buffer_manager.h
```diff
@@ -202,12 +301,102 @@ class WriteBufferManager final {
...
+  // Assumed the lock is held
+  void WakeupFlushInitiationThread() {
+    new_flushes_wakeup_ = true;
+    flushes_wakeup_cv_.notify_one();
+  }
+
+  // This is used outside the flushes_mu_ lock => only
+  // additional_flush_initiation_size_ needs to be atomic
+  bool ShouldInitiateAnotherFlushMemOnly(size_t curr_memory_used) const {
+    return (curr_memory_used >= additional_flush_initiation_size_);
```
isaac-io, line 342: Don't we want to return true only for the time that this threshold has been crossed rather than letting many writers initiate flushes until the flush limit is reached? With `curr_memory_used` alone we don't have enough information to decide, but with the size of the memory before and after the change we'll be able to do that.

include/rocksdb/write_buffer_manager.h
```diff
@@ -202,12 +301,102 @@ class WriteBufferManager final {
...
+  void WakeUpFlushesThread();
+  void TerminateFlushesThread();
+  void RecalcFlushInitiationSize();
+  void ReevaluateNeedForMoreFlushes(size_t curr_memory_used);
+  uint64_t FindInitiator(void* initiator) const;
+
+  // Assumed the lock is held
+  void WakeupFlushInitiationThread() {
+    new_flushes_wakeup_ = true;
+    flushes_wakeup_cv_.notify_one();
```
isaac-io, line 336: Doing the notification under the lock is unnecessarily expensive, because the notified thread will wake up and try to acquire the mutex, only to be put back to sleep until this thread releases it.

include/rocksdb/write_buffer_manager.h
```diff
@@ -202,12 +301,102 @@ class WriteBufferManager final {
...
+  void InitiateFlushesThread();
+  bool InitiateAdditionalFlush();
+  void WakeUpFlushesThread();
+  void TerminateFlushesThread();
+  void RecalcFlushInitiationSize();
+  void ReevaluateNeedForMoreFlushes(size_t curr_memory_used);
+  uint64_t FindInitiator(void* initiator) const;
+
+  // Assumed the lock is held
+  void WakeupFlushInitiationThread() {
```
isaac-io, line 334: This assumption isn't actually honoured in `RegisterFlushInitiator()` and `DeregisterFlushInitiator()`.

memtable/write_buffer_manager.cc
```diff
@@ -243,7 +291,382 @@ std::string WriteBufferManager::GetPrintableOptions() const {
...
+    // TODO - Update this to a formula. If it depends on the number of initators
+    // => update when that number changes
+    min_flush_size_ = 4 * 1024U * 1024U;  // 4 MB
+    RecalcFlushInitiationSize();
+  }
+
+  if (flushes_thread_.joinable() == false) {
+    flushes_thread_ =
+        std::thread(&WriteBufferManager::InitiateFlushesThread, this);
+  }
```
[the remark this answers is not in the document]
udi-speedb: Because the thread is only initialized if the flush initiation mechanism is enabled and after proper initialization of the applicable mechanism variables.

memtable/write_buffer_manager.cc
udi-speedb: The flushes thread releases the main flushes lock (flushes_mu_) when invoking the callback to initiate flushes. So, how would registration / deregistration know that the thread has completed a cycle?

ayulas:
i left some comments. part of isaac i also thinks but i. didnt repeat

udi-speedb:
> i left some comments. part of isaac i also thinks but i. didnt repeat

@ayulas  - I do not see any comment you have made.

udi-speedb:
@hilikspdb - I have added you as a reviewer as well as you are the original designer of the feature and your inputs would be helpful

db/db_impl/db_impl_compaction_flush.cc
```diff
@@ -3786,4 +3795,100 @@ Status DBImpl::WaitForCompact(bool wait_unscheduled) {
...
+  // mutable memtable as well. However, here we reuest a flush of only the
+  // immutable ones => the freed memory by flushing is not the same as the one
+  // causing this cf to be picked
+  FlushOptions flush_options;
+  flush_options.allow_write_stall = true;
+  flush_options.wait = false;
+  Status s;
+  if (immutable_db_options_.atomic_flush) {
+    s = AtomicFlushMemTables({cfd_to_flush}, flush_options,
+                             FlushReason::kSpeedbWriteBufferManager);
```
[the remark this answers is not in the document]
udi-speedb: I have created a separate function per type of flushing (atomic / not atomic). The atomic-flush code now sends all of the cf-s selected for atomic flush for flushing.

include/rocksdb/listener.h
```diff
@@ -169,6 +169,7 @@ enum class FlushReason : int {
   // will not be called to avoid many small immutable memtables.
   kErrorRecoveryRetryFlush = 0xc,
   kWalFull = 0xd,
+  kSpeedbWriteBufferManager = 0xe,
```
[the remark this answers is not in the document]
udi-speedb: Leaving it for now. Indeed there should be either the existing WBM flushes or WBM-Initiated flushing, but I prefer having a unique constant (renamed to kWriteBufferManagerInitiated) so we clearly see the type in the log.

include/rocksdb/write_buffer_manager.h
udi-speedb: Code was modified so the notification is not under the lock anymore.

include/rocksdb/write_buffer_manager.h
udi-speedb: All notifications are not under the lock now

db/db_impl/db_impl_compaction_flush.cc
udi-speedb: The motivation is to avoid starvation of a cf and balance the need to flush cf-s to release their WAL-s.

memtable/write_buffer_manager.cc
udi-speedb: Checking a DB requires locking the db's mutex. Also, by doing round-robin, avoids starving a DB.

memtable/write_buffer_manager.cc
udi-speedb: Assertion replaced with an if and counter decreased only if it is > 0

memtable/write_buffer_manager.cc
udi-speedb: Function removed altogether. It's irrelevant without the code of the delays (#114 )

memtable/write_buffer_manager.cc
```diff
@@ -247,7 +289,273 @@ std::string WriteBufferManager::GetPrintableOptions() const {
...
+      auto num_repeated_failures_to_initiate = 0U;
+      while ((num_repeated_failures_to_initiate < num_initiators_) &&
+             (num_flushes_to_initiate_ > 0U)) {
+        bool was_flush_initiated = false;
+        // Unlocking the flushed_mu_ since flushing (via the initiator cb) may
+        // call a WBM service (e.g., ReserveMem()), that, in turn, needs to
+        // flushes_mu_lock the same mutex => will get stuck
+        // // flushes_mu_lock.unlock();
+        {
+          InstrumentedMutexUnlock flushes_mu_unlocker(flushes_mu_.get());
```
ayulas, line 411: i think i missed something.... or it is a problem... you release here the flushes_mu_ that was taken in the InstrumentedMutexLock lock(flushes_mu_.get()) wait for work. but you are in loop and i dont see you acquire it again... so you unlock on non lock??? i probably miss something.

memtable/write_buffer_manager.cc
```diff
@@ -247,7 +289,273 @@ std::string WriteBufferManager::GetPrintableOptions() const {
...
+void WriteBufferManager::InitiateFlushesThread() {
+  while (true) {
+    // Should return true when the waiting should stop (no spurious wakeups
+    // guaranteed)
+    auto StopWaiting = [this]() {
+      return (new_flushes_wakeup_ &&
+              (terminate_flushes_thread_ || (num_flushes_to_initiate_ > 0U)));
+    };
+
+    // // std::unique_lock<std::mutex> flushes_mu_lock(*flushes_mu_);
```
ayulas, line 364: why not using CV?

memtable/write_buffer_manager.cc
```diff
@@ -247,7 +289,273 @@ std::string WriteBufferManager::GetPrintableOptions() const {
...
 }
 
+// ================================================================================================
+void WriteBufferManager::RegisterFlushInitiator(
+    void* initiator, InitiateFlushRequestCb request) {
+  {
+    std::unique_lock<std::mutex> lock(flushes_initiators_mu_);
+    assert(IsInitiatorIdxValid(FindInitiator(initiator)) == false);
+    flush_initiators_.push_back({initiator, request});
+    num_initiators_ = flush_initiators_.size();
```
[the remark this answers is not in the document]
udi-speedb: - num_initiators_ is itself atomic as it is read and written in different threads
- Modifying it is always under the same lock (flushes_initiators_mu_ and it's atomic anyway). What issue do you see with accessing it? What
- Taking the flushes_initiators_mu_ lock just to access flush_initiators_.size() seems like a high price to pay for that

memtable/write_buffer_manager.cc
udi-speedb: I could use an unordered_map. However, since I am doing a round robin of the initiators and wish to cycle through them in-order, it is easier to guarantee when they register / de-register (also, I have an existing unit test with many tests that rely on that and will have to more or less be rewritten if I move to an unordered container).
Registration / de-registration is rare, so the price for the vector operations is negligible

memtable/write_buffer_manager.cc
```diff
@@ -59,22 +72,35 @@ std::size_t WriteBufferManager::dummy_entries_in_cache_usage() const {
...
+  auto is_enabled = enabled();
+  size_t new_memory_used = 0U;
+
   if (cache_res_mgr_ != nullptr) {
-    ReserveMemWithCache(mem);
-  } else if (enabled()) {
-    memory_used_.fetch_add(mem, std::memory_order_relaxed);
+    new_memory_used = ReserveMemWithCache(mem);
+  } else if (is_enabled) {
+    auto old_memory_used =
```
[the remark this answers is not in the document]
udi-speedb: I would like to have the new memory in a local variable rather than access the atomic variable repeatedly. Also, putting it in a local variable guarantees that its value is consistent in all the places that access it in a single function call.

memtable/write_buffer_manager.cc
```diff
@@ -59,22 +72,35 @@ std::size_t WriteBufferManager::dummy_entries_in_cache_usage() const {
...
-void WriteBufferManager::ReserveMemWithCache(size_t mem) {
+size_t WriteBufferManager::ReserveMemWithCache(size_t mem) {
 #ifndef ROCKSDB_LITE
   assert(cache_res_mgr_ != nullptr);
   // Use a mutex to protect various data structures. Can be optimized to a
   // lock-free solution if it ends up with a performance bottleneck.
   std::lock_guard<std::mutex> lock(cache_res_mgr_mu_);
 
-  size_t new_mem_used = memory_used_.load(std::memory_order_relaxed) + mem;
+  size_t old_mem_used = memory_used_.load(std::memory_order_relaxed);
```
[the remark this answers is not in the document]
udi-speedb: Same reason as above

memtable/write_buffer_manager.cc
```diff
@@ -113,13 +142,15 @@ void WriteBufferManager::FreeMemAborted(size_t mem) {
...
 void WriteBufferManager::FreeMem(size_t mem) {
   const auto is_enabled = enabled();
+  size_t new_memory_used = 0U;
 
   if (cache_res_mgr_ != nullptr) {
-    FreeMemWithCache(mem);
+    new_memory_used = FreeMemWithCache(mem);
   } else if (is_enabled) {
-    [[maybe_unused]] const auto curr_memory_used =
+    auto old_memory_used =
```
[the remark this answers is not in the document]
udi-speedb: Same

memtable/write_buffer_manager.cc
```diff
@@ -247,7 +289,273 @@ std::string WriteBufferManager::GetPrintableOptions() const {
...
+}
+
+void WriteBufferManager::DeregisterFlushInitiator(void* initiator) {
+  {
+    std::unique_lock<std::mutex> lock(flushes_initiators_mu_);
+    auto initiator_idx = FindInitiator(initiator);
+    assert(IsInitiatorIdxValid(initiator_idx));
+
+    flush_initiators_.erase(flush_initiators_.begin() + initiator_idx);
+    num_initiators_ = flush_initiators_.size();
```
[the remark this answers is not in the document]
udi-speedb: Same reply. 
Updating num_initiators_ under the flushes_initiators_mu_ lock and it is an std::atomic. I have yet to understand the issue here

memtable/write_buffer_manager.cc
```diff
@@ -247,7 +289,273 @@ std::string WriteBufferManager::GetPrintableOptions() const {
...
+          auto& initiator = flush_initiators_[next_candidate_initiator_idx_];
+          next_candidate_initiator_idx_ = CalcNextCandidateInitiatorIdx();
+
+          // Assuming initiator would flush (flushes_mu_lock is unlocked and
+          // called initiator may call another method that relies on these
+          // counters) Not recalculating flush initiation size since the
+          // increment & decrement cancel each other with respect to the recalc
+          ++num_running_flushes_;
+          --num_flushes_to_initiate_;
+          was_flush_initiated = initiator.cb(kMinFlushSizes[iter]);
```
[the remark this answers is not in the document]
udi-speedb: Good point. I have updated the heuristics to account for this:
- I have added a counter to the ColumnFamilyData for the number of times it has been sent to flush.
- When picking a CF, if its counter is lagging behind too much (> threshold) relative to the other CF-s, I will pick it, regardless of the size of the data to flush it has.
You will see the code in the next commit and I am looking forward to your review and comments.

memtable/write_buffer_manager.cc
udi-speedb: This is an InstrumentedMutexUnlock (an unlocker). It's destructor (re)locks the mutex. It is part of a scoped block => It will be destructed and reacquire the mutex.

memtable/write_buffer_manager.cc
udi-speedb: I didn't understand this comment.

udi-speedb:
@ayulas - Thanks a lot for the review.

memtable/write_buffer_manager.cc
```diff
@@ -247,7 +290,304 @@ std::string WriteBufferManager::GetPrintableOptions() const {
...
 
+  snprintf(buffer, kBufferSize, "%*s: %d\n", field_width,
+           "wbm.initiate_flushes", IsInitiatingFlushes());
+  ret.append(buffer);
+
   return ret;
 }
 
+// ================================================================================================
+void WriteBufferManager::RegisterFlushInitiator(
```
ayulas, line 301: why the registration object is not a smart pointer? it will help you handle the fact the the object was deleted and by that not need protect by the lock....
the initiator object needs to be an element in the parent object (or inherit from an initiator class ) -   the consist object is a generic, right?  - currently is the DB but tomorrow it can be something else.

memtable/write_buffer_manager.cc
```diff
@@ -247,7 +290,304 @@ std::string WriteBufferManager::GetPrintableOptions() const {
...
+  // that were running when it was enabled will have completed.
+  if (num_running_flushes_ > 0U) {
+    --num_running_flushes_;
+  }
+  size_t curr_memory_used = memory_usage();
+  RecalcFlushInitiationSize();
+  ReevaluateNeedForMoreFlushesLockHeld(curr_memory_used);
+}
+
+void WriteBufferManager::RecalcFlushInitiationSize() {
```
ayulas, line 505: please do assert if the flushes_mu_ is not held. this function must be called under this flushes_mu_ lock

memtable/write_buffer_manager.cc
ayulas: my bad you are correct

memtable/write_buffer_manager.cc
```diff
@@ -247,7 +290,304 @@ std::string WriteBufferManager::GetPrintableOptions() const {
...
+void WriteBufferManager::InitiateFlushesThread() {
+  while (true) {
+    // Should return true when the waiting should stop (no spurious wakeups
+    // guaranteed)
+    auto StopWaiting = [this]() {
+      return (new_flushes_wakeup_ &&
+              (terminate_flushes_thread_ || (num_flushes_to_initiate_ > 0U)));
+    };
+
+    InstrumentedMutexLock lock(flushes_mu_.get());
```
ayulas, line 372: why not using the common format of CV?
    std::unique_lock<std::mutex> 
    flushes_wakeup_cv_.wait ?

memtable/write_buffer_manager.cc
```diff
@@ -247,7 +290,304 @@ std::string WriteBufferManager::GetPrintableOptions() const {
...
+    // 1. Flushes of a min size.
+    // 2. Flushes of any size
+    constexpr size_t kNumIters = 2U;
+    const std::array<size_t, kNumIters> kMinFlushSizes{min_mutable_flush_size_,
+                                                       0U};
+
+    auto iter = 0U;
+    while ((iter < kMinFlushSizes.size()) && (num_flushes_to_initiate_ > 0U)) {
+      auto num_repeated_failures_to_initiate = 0U;
+      while ((num_repeated_failures_to_initiate < num_initiators_) &&
```
ayulas, line 409: here is a bug accessing num_initiators_ that is bigger than the initiators containers size. its not terrible but there is a  better option

      while (num_flushes_to_initiate_ > 0U) {
        bool was_flush_initiated = false;
        // Unlocking the flushed_mu_ since flushing (via the initiator cb) may
        // call a WBM service (e.g., ReserveMem()), that, in turn, needs to
        // flushes_mu_lock the same mutex => will get stuck
        // // flushes_mu_lock.unlock();
        {
          InstrumentedMutexUnlock flushes_mu_unlocker(flushes_mu_.get());

          std::unique_lock<std::mutex> initiators_lock(flushes_initiators_mu_);
          if (num_repeated_failures_to_initiate >= flush_initiators_.size()) { 
              break;
          }
         ....

memtable/write_buffer_manager.cc
```diff
@@ -247,7 +290,304 @@ std::string WriteBufferManager::GetPrintableOptions() const {
...
+          auto& initiator = flush_initiators_[next_candidate_initiator_idx_];
+          UpdateNextCandidateInitiatorIdx();
+
+          // Assuming initiator would flush (flushes_mu_lock is unlocked and
+          // called initiator may call another method that relies on these
+          // counters) Not recalculating flush initiation size since the
+          // increment & decrement cancel each other with respect to the recalc
+          ++num_running_flushes_;
+          --num_flushes_to_initiate_;
+          was_flush_initiated = initiator.cb(kMinFlushSizes[iter]);
```
ayulas, line 436: as i mention the initiator should be a smart pointer so no need protect the flush in the flushes_initiators_mu_ scope

memtable/write_buffer_manager.cc
```diff
@@ -247,7 +290,304 @@ std::string WriteBufferManager::GetPrintableOptions() const {
...
+          if (!was_flush_initiated) {
+            // No flush was initiated => undo the counters update
+            --num_running_flushes_;
+            ++num_flushes_to_initiate_;
+          }
+        }
+
+        if (was_flush_initiated) {
```
ayulas, line 444: why you do it not under line 437?

memtable/write_buffer_manager.cc
```diff
@@ -9,26 +9,35 @@
...
       memory_used_(0),
       memory_inactive_(0),
       memory_being_freed_(0U),
       cache_res_mgr_(nullptr),
       allow_stall_(allow_stall),
-      stall_active_(false) {
+      stall_active_(false),
+      initiate_flushes_(initiate_flushes),
+      flush_initiation_options_(flush_initiation_options),
+      flushes_mu_(new InstrumentedMutex),
```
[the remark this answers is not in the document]
udi-speedb: including instrumented_mutex.h results causes compilation to fail.
I have opened an issue about it (#279)

memtable/write_buffer_manager.cc
```diff
@@ -247,7 +290,304 @@ std::string WriteBufferManager::GetPrintableOptions() const {
...
+  ret.append(buffer);
+
   return ret;
 }
 
+// ================================================================================================
+void WriteBufferManager::RegisterFlushInitiator(
+    void* initiator, InitiateFlushRequestCb request) {
+  {
+    InstrumentedMutexLock lock(flushes_initiators_mu_.get());
```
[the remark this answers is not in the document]
udi-speedb: See above

memtable/write_buffer_manager.cc
udi-speedb: I suggest we discuss that over a zoom session. I didn't understand your comment.

memtable/write_buffer_manager.cc
udi-speedb: Added flushes_mu_->AssertHeld()

memtable/write_buffer_manager.cc
udi-speedb: I do not understand the comment. Let's discuss over a zoom session

memtable/write_buffer_manager.cc
udi-speedb: I do not understand the comment.
The flushes_initiator_mu_ protects the access to the vector containing the initiators.
Let's discuss over a zoom session

db/db_impl/db_impl_compaction_flush.cc
```diff
@@ -3894,6 +3894,10 @@ bool DBImpl::InitiateMemoryManagerFlushRequestAtomicFlush(
     }
   }
 
+  if (cfds.empty()) {
```
ayulas, line 3897: actually you should be protected with the 
    if (total_size_to_flush < min_size_to_flush) {
      return false;
    }
cause in case no cfds in DB the total_size_to_flush should be 0 so it menas you passed min_size_to_flush 0 and this is a bug... (not related to this )
second i would have cgecked this after taking the InstrumentedMutexLock

udi-speedb:
@Yuval-Ariel - Please run full QA on this branch. Thanks

db/db_impl/db_impl_compaction_flush.cc
```diff
@@ -3843,4 +3853,167 @@ Status DBImpl::WaitForCompact(bool wait_unscheduled) {
...
+
+bool DBImpl::InitiateMemoryManagerFlushRequestAtomicFlush(
+    size_t min_size_to_flush, const FlushOptions& flush_options) {
+  assert(immutable_db_options_.atomic_flush);
+
+  autovector<ColumnFamilyData*> cfds;
+  {
+    InstrumentedMutexLock lock(&mutex_);
+    SelectColumnFamiliesForAtomicFlush(&cfds);
+
```
ayulas, line 3882: why not check if cfds is empty here and return? why move on?
also as i mention in previous review comment - eventhough you the cfds is not empty if all cfds are with no data you should have exit in total_size_to_flush < min_size_to_flush. so if you needed protect this you have a bug that min_size_to_flush is 0

db/db_impl/db_impl_compaction_flush.cc
```diff
@@ -3843,4 +3853,167 @@ Status DBImpl::WaitForCompact(bool wait_unscheduled) {
...
+  // The CF picked based on min min_size_to_flush
+  ColumnFamilyData* orig_cfd_to_flush = nullptr;
+  // The cf to actually flush (possibly == orig_cfd_to_flush)
+  ColumnFamilyData* cfd_to_flush = nullptr;
+  SequenceNumber seq_num_for_cf_picked = kMaxSequenceNumber;
+
+  {
+    InstrumentedMutexLock lock(&mutex_);
+
```
ayulas, line 3942: again exit immediately if cfds is empty...

db/db_impl/db_impl_compaction_flush.cc
```diff
@@ -3843,4 +3853,167 @@ Status DBImpl::WaitForCompact(bool wait_unscheduled) {
...
+    }
+
+    if (cfd_to_flush == nullptr) {
+      return false;
+    }
+
+    orig_cfd_to_flush = cfd_to_flush;
+
+    // A CF was picked. Now see if it should be replaced with a lagging CF
+    for (auto* cfd : *versions_->GetColumnFamilySet()) {
```
ayulas, line 3967: why you created 2 loops?? you can do it in 1

udi-speedb:
@ayulas - I have pushed the changes we have discussed over zoom.
Please take a look and approve the PR so @Yuval-Ariel may start QA.
@yuval - Before you start the QA, please let me know and I will both rebase on latest origin/main and squash all of the commits. Thanks

ayulas:
i approve. @Yuval-Ariel go ahead

Yuval-Ariel (approved):

PR #191, opened 2022-10-03
CI: run stress tests for 1hr instead of 500secs

### LINKED ISSUE #190: stress tests are too short
500 secs is not enough. change to 1hr.

Yuval-Ariel:
closes #190

### DIFF
--- a/.github/workflows/qa-tests.yml
+++ b/.github/workflows/qa-tests.yml
@@ -24,10 +24,10 @@ jobs:
             short_test: TMPD="$(mktemp -d /tmp/speedb.XXXX)" make -j$(nproc) check
             long_test: TMPD="$(mktemp -d /tmp/speedb.XXXX)" make -j$(nproc) check
           - name: black-box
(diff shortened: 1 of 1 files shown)


isaac-io (approved):
LGTM. Is there a reason to choose 1 hour specifically? Currently CI runs this on every PR. Ideally we would want to avoid running this when there are no code changes, but even then, trivial code changes waiting an hour for stress testing to finish (and having to go through it all when rebasing, etc.), seems a bit much.

Yuval-Ariel:
> LGTM. Is there a reason to choose 1 hour specifically? Currently CI runs this on every PR. Ideally we would want to avoid running this when there are no code changes, but even then, trivial code changes waiting an hour for stress testing to finish (and having to go through it all when rebasing, etc.), seems a bit much.

trivial code changes shouldnt have to go through stress and unit tests right? the problem right now is that the stress is not long enough for actual code changes.. we could add an additional pipeline which will be longer or maybe its possible to tune it with some external arg?

Issue #169, opened 2022-09-21
Change default compaction threads from 1 to 8


bosmatt:
Changing the default compaction threads shows performance improvement as can be seen below.
 We believe that even there are not too many cores in the system, the performance is still better with more threads than with a single thread. 

![Image](https://user-images.githubusercontent.com/107058910/191572883-55ed2004-5320-42f7-8a3d-87ac840bee6b.png)

erez-speedb:
@isaac-io  testing is done, please merge

isaac-io:
Thanks! Waiting for Assaf to approve that QA tests passed as well.

PR #194, opened 2022-10-19
options: change the default value for compaction threads to 8 (#169)

isaac-io:
This is done through the deprecated `max_background_compactions` option, which overrides `max_background_jobs`, so in order to avoid having an adverse effect on users who set `max_background_jobs` to a high value we try to detect that case and shift back to the user's choice in case there are enough background jobs to be enough for both flushes and compactions.

Test plan: run `make check` with the fixed unit tests.

### DIFF
--- a/db/column_family_test.cc
+++ b/db/column_family_test.cc
@@ -3128,8 +3128,6 @@ TEST_P(ColumnFamilyTest, IteratorCloseWALFile2) {
 #ifndef ROCKSDB_LITE  // TEST functions are not supported in lite
 TEST_P(ColumnFamilyTest, ForwardIteratorCloseWALFile) {
   SpecialEnv env(Env::Default());
-  // Allow both of flush and purge job to schedule.
-  env.SetBackgroundThreads(2, Env::HIGH);
   db_options_.env = &env;
   db_options_.max_background_flushes = 1;
   column_family_options_.memtable_factory.reset(
@@ -3163,9 +3161,8 @@ TEST_P(ColumnFamilyTest, ForwardIteratorCloseWALFile) {
   ROCKSDB_NAMESPACE::SyncPoint::GetInstance()->LoadDependency({
       {"ColumnFamilyTest::IteratorCloseWALFile2:0",
        "DBImpl::BGWorkPurge:start"},
-      {"ColumnFamilyTest::IteratorCloseWALFile2:2",
+      {"ColumnFamilyTest::IteratorCloseWALFile2:1",
        "DBImpl::BackgroundCallFlush:start"},
-      {"DBImpl::BGWorkPurge:end", "ColumnFamilyTest::IteratorCloseWALFile2:1"},
   });
   ROCKSDB_NAMESPACE::SyncPoint::GetInstance()->EnableProcessing();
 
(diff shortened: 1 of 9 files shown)


db/column_family_test.cc
```diff
@@ -3128,8 +3128,6 @@ TEST_P(ColumnFamilyTest, IteratorCloseWALFile2) {
 #ifndef ROCKSDB_LITE  // TEST functions are not supported in lite
 TEST_P(ColumnFamilyTest, ForwardIteratorCloseWALFile) {
```
Yuval-Ariel, line 3129: can you describe the problem and explain your fix?

isaac-io: Sure. The problem is that the `Seek()` call on the tailing iterator causes two purges to be scheduled, and when there was only a single background thread they were scheduled one after the other, with the first one closing and removing the WAL file and the second one doing nothing. With 8 threads available, the purges are scheduled concurrently, causing the original `ColumnFamilyTest::IteratorCloseWALFile2:1` sync point to be released by the purge that does nothing before the real purge finishes closing and removing the WAL file, which is why the test was failing.

The fix is to fill the low priority pool with sleeping tasks after the purges have been scheduled. When all of the tasks begin sleeping, that's an indication that they have been scheduled, meaning that the purges are done and we can just free the pool and move on to checking the amount of open WAL files.

db/db_wal_test.cc
```diff
@@ -1496,6 +1496,9 @@ TEST_F(DBWALTest, RaceInstallFlushResultsWithWalObsoletion) {
   options.track_and_verify_wals_in_manifest = true;
   // The following make sure there are two bg flush threads.
   options.max_background_jobs = 8;
+  options.max_background_compactions = options.max_background_flushes = -1;
+  env_->SetBackgroundThreads(1, Env::Priority::HIGH);
+  env_->SetBackgroundThreads(1, Env::Priority::LOW);
```
Yuval-Ariel, line 1501: i dont think this is required since just after CreateAndReopenWithCF is called which resizes the threadpool according to options passed

db/db_options_test.cc
```diff
@@ -615,6 +615,9 @@ TEST_F(DBOptionsTest, SetBackgroundJobs) {
   Options options;
   options.create_if_missing = true;
   options.max_background_jobs = 8;
+  options.max_background_compactions = options.max_background_flushes = -1;
+  env_->SetBackgroundThreads(1, Env::Priority::HIGH);
+  env_->SetBackgroundThreads(1, Env::Priority::LOW);
```
Yuval-Ariel, line 620: i dont think these are required since db is reopened just after with 6 LOW and 2 HIGH in Reopen(options);

db/db_wal_test.cc
isaac-io: It is needed because the DB is initially opened before the test begins, and since the default is now 8 compaction jobs, it increases the low priority thread pool to have 8 threads in it, and reopening with a smaller amount doesn't reduce it (because `DBImpl::Open()` is calling `IncBackgroundThreadsIfNeeded()`, not `SetBackgroundThreads()`, so it can only increase the amount that's already there), so we'll end up with effectively 10 background jobs (2 in high, and the previous 8 in low) rather than 8 (2 in high and 6 in low) as expected. That's why I force a pool reduction before reopening with these options.

db/db_options_test.cc
isaac-io: Same as in the other comment: It is needed because the DB is initially opened before the test begins, and since the default is now 8 compaction jobs, it increases the low priority thread pool to have 8 threads in it, and reopening with a smaller amount doesn't reduce it (because `DBImpl::Open()` is calling `IncBackgroundThreadsIfNeeded()`, not `SetBackgroundThreads()`, so it can only increase the amount that's already there), so we'll end up with effectively 10 background jobs (2 in high, and the previous 8 in low) rather than 8 (2 in high and 6 in low) as expected. That's why I force a pool reduction before reopening with these options.

Yuval-Ariel (approved):

Issue #197, opened 2022-10-23
Unit tests intermittently failing on `main`


isaac-io:
The following tests sometimes fail on `main`:

```
$ ./db_compaction_test --gtest_filter=DBCompactionTestWithParam/DBCompactionTestWithParam.PartialCompactionFailure/1
Note: Google Test filter = DBCompactionTestWithParam/DBCompactionTestWithParam.PartialCompactionFailure/1
[==========] Running 1 test from 1 test case.
[----------] Global test environment set-up.
[----------] 1 test from DBCompactionTestWithParam/DBCompactionTestWithParam
[ RUN      ] DBCompactionTestWithParam/DBCompactionTestWithParam.PartialCompactionFailure/1
/home/isaac/projects/speedb-io/db/db_compaction_test.cc:2894: Failure
Value of: !dbfull()->TEST_WaitForCompact().ok()
  Actual: false
Expected: true
8:0
[  FAILED  ] DBCompactionTestWithParam/DBCompactionTestWithParam.PartialCompactionFailure/1, where GetParam() = (1, false) (290 ms)
[----------] 1 test from DBCompactionTestWithParam/DBCompactionTestWithParam (290 ms total)

[----------] Global test environment tear-down
[==========] 1 test from 1 test case ran. (290 ms total)
[  PASSED  ] 0 tests.
[  FAILED  ] 1 test, listed below:
[  FAILED  ] DBCompactionTestWithParam/DBCompactionTestWithParam.PartialCompactionFailure/1, where GetParam() = (1, false)

 1 FAILED TEST
```

```
$ ./deletefile_test --gtest_filter=DeleteFileTest.BackgroundPurgeCopyOptions
Note: Google Test filter = DeleteFileTest.BackgroundPurgeCopyOptions
[==========] Running 1 test from 1 test case.
[----------] Global test environment set-up.
[----------] 1 test from DeleteFileTest
[ RUN      ] DeleteFileTest.BackgroundPurgeCopyOptions
/home/isaac/projects/speedb-io/db/deletefile_test.cc:124: Failure
Expected equality of these values:
  required_sst
    Which is: 1
  sst_cnt
    Which is: 2
[  FAILED  ] DeleteFileTest.BackgroundPurgeCopyOptions (9732 ms)
[----------] 1 test from DeleteFileTest (9732 ms total)

[----------] Global test environment tear-down
[==========] 1 test from 1 test case ran. (9734 ms total)
[  PASSED  ] 0 tests.
[...]

PR #198, opened 2022-10-23
tests: fix breakage caused by the compaction threads value change (#197)

isaac-io:
In #194 the default value for background compaction threads was changed to 8 (from 1). This caused some tests to fail, but they were fixed. Alas, the fixes seems to not be complete, as under stress some tests still fail.

Make it so the tests are truly fixed now. This can be checked by running them in a loop with the machine overloaded:

```
$ while ./db_compaction_test --gtest_filter=DBCompactionTestWithParam/DBCompactionTestWithParam.PartialCompactionFailure/1; do sleep 1; done
```

And

```
$ while ./deletefile_test --gtest_filter=DeleteFileTest.BackgroundPurgeCopyOptions; do sleep 1; done
```

### DIFF
--- a/db/db_compaction_test.cc
+++ b/db/db_compaction_test.cc
@@ -2848,14 +2848,16 @@ TEST_P(DBCompactionTestWithParam, PartialCompactionFailure) {
   options.max_bytes_for_level_multiplier = 2;
   options.compression = kNoCompression;
   options.max_subcompactions = max_subcompactions_;
-  options.max_background_compactions = 1;
 
   env_->SetBackgroundThreads(1, Env::HIGH);
-  env_->SetBackgroundThreads(1, Env::LOW);
   // stop the compaction thread until we simulate the file creation failure.
-  test::SleepingBackgroundTask sleeping_task_low;
-  env_->Schedule(&test::SleepingBackgroundTask::DoSleepTask, &sleeping_task_low,
-                 Env::Priority::LOW);
+  std::vector<test::SleepingBackgroundTask> sleeping_task_low(
+      std::max(1, env_->GetBackgroundThreads(Env::Priority::LOW)));
+  for (auto& sleeping_task : sleeping_task_low) {
+    env_->Schedule(&test::SleepingBackgroundTask::DoSleepTask, &sleeping_task,
+                   Env::Priority::LOW);
+    sleeping_task.WaitUntilSleeping();
+  }
 
   options.env = env_;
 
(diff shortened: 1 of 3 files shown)


Yuval-Ariel (approved):
LGTM

Yuval-Ariel:
CI failed, looks like theres another such test :
2122	:	1666534254.761	     0.105	0	803	1	0	./t/run-db_range_del_test-DBRangeDelTest.TableEvictedDuringScan >& t/log-run-db_range_del_test-DBRangeDelTest.TableEvictedDuringScan || bash -c "cat t/log-run-db_range_del_test-DBRangeDelTest.TableEvictedDuringScan; exit $?"

but this seems more related to the setting of purges as low priority. iterator deletion timing?

isaac-io:
> CI failed, looks like theres another such test : 2122 : 1666534254.761 0.105 0 803 1 0 ./t/run-db_range_del_test-DBRangeDelTest.TableEvictedDuringScan >& t/log-run-db_range_del_test-DBRangeDelTest.TableEvictedDuringScan || bash -c "cat t/log-run-db_range_del_test-DBRangeDelTest.TableEvictedDuringScan; exit $?"
> 
> but this seems more related to the setting of purges as low priority. iterator deletion timing?

Doesn't seem to be the case. The iterator is not created with background purge, and in any case I don't see why it would get stuck because of purges. There's nothing blocking them. Also, this test is about snapshots, and that should not block purges (it only uses the iterator after the compaction finished, so there shouldn't be anything to purge anyway).

I'm also unable to reproduce it locally, even under stress, so this seems like a spurious error. I'll trigger the unit tests run again just to be sure.

Yuval-Ariel:
heres the error:
[ RUN      ] DBRangeDelTest.TableEvictedDuringScan
db/db_range_del_test.cc:659: Failure
Expected: (NumTableFilesAtLevel(1)) > (1), actual: 1 vs 1
[  FAILED  ] DBRangeDelTest.TableEvictedDuringScan (45 ms)
seems unrelated to this change so once the CI passes its good to go

isaac-io:
> heres the error: [ RUN ] DBRangeDelTest.TableEvictedDuringScan db/db_range_del_test.cc:659: Failure Expected: (NumTableFilesAtLevel(1)) > (1), actual: 1 vs 1 [ FAILED ] DBRangeDelTest.TableEvictedDuringScan (45 ms) seems unrelated to this change so once the CI passes its good to go

Oh. That's actually probably caused by the changed in #194 as well. The test has two levels (L0 and L1) with compaction trigger and stop write trigger set to 4. It then creates 25 L0 files, which should compact to 1 or more L1 files. However, the test require there to be more than a single file in L1, and here timing matters, because it might be that the call to `TEST_WaitForCompact()` is delayed enough to catch a second concurrently scheduled compaction, causing it to arrive at a state where only a single file is left at L1. Nice catch! I'll push a fix shortly.

Yuval-Ariel (approved):

isaac-io:
Unit tests succeeded: https://github.com/speedb-io/speedb/actions/runs/3310563114/jobs/5464918153

Issue #231, opened 2022-11-10
version: the build tag is incorrectly added to the RocksDB version string


isaac-io:
The changes in #156 were accidentally applied to the `GetRocksVersionAsString()` function instead of `GetSpeedbVersionAsString()`. This replaced `ROCKSDB_PATCH` with `SPEEDB_PATCH` (causing 7.2.2 to become 7.2.0), as well as added the tag in the wrong place (`Speedb version: 2.1.0 (7.2.0-*-(main+4)-(#isaac/snapshot+1))` instead of `Speedb version: 2.1.0-*-(main+4)-(#isaac/snapshot+1) (7.2.2)`).

## To Reproduce
1. Build `db_bench` on any branch containing #156 
2. Run it and see the log file

## Expected behavior
The RocksDB version patch isn't replaced, and the tag is applied to the Speedb version rather than the RocksDB one.

## Additional context
N/A

isaac-io:
@erez-speedb does this means that you'll need to make changes to the way the tag is extracted and shown in the admin site?

PR #233, opened 2022-11-10
build_version: apply the build tag to the Speedb version string (#231)

isaac-io:
The changes in #157 were accidentally applied to the `GetRocksVersionAsString()` function instead of the `GetSpeedbVersionAsString()` function. This replaced the RocksDB patch number with the Speedb one, and added the build tag in the wrong place.

Fix it by moving the logic to the intended function.


mrambacher (approved):
Thanks for the quick fix!

Yuval-Ariel (approved):

PR #244, opened 2022-11-15
120 qa tool for testing the artifacts

RoyBenMoshe:


tools/artifacts_check/Makefile
```diff
@@ -0,0 +1,29 @@
...
+endif
+
+CFLAGS += -Wstrict-prototypes
+
+.PHONY: clean
+
+all: check_static check_shared
+
+check_static: check_artifacts.cc
+	$(CXX) $(CXXFLAGS) check_artifacts.cc -o$@ ../../libspeedb.a -I../../include -O2 -std=c++17 $(PLATFORM_LDFLAGS) $(PLATFORM_CXXFLAGS) $(EXEC_LDFLAGS)
```
Yuval-Ariel, line 22: the -O2 optimize level is used only when DEBUG_LEVEL=2. why are we using it here?

Yuval-Ariel (approved):

PR #257, opened 2022-11-21
221 set dbs to use per benchmark

udi-speedb:


tools/db_bench_tool.cc
```diff
@@ -6833,7 +6870,7 @@ class Benchmark {
         ErrorExit("put or merge error: %s", s.ToString().c_str());
       }
       bytes += key.size() + val.size() + user_timestamp_size_;
-      thread->stats.FinishedOps(&db_, db_.db, 1, kWrite);
+      thread->stats.FinishedOps(&SingleDb(), SingleDb().db, 1, kWrite);
```
Yuval-Ariel, line 6873: same as "FinishedOps comment"

tools/db_bench_tool.cc
```diff
@@ -6666,7 +6703,7 @@ class Benchmark {
         thread->stats.ResetLastOpTime();
       }
 
-      thread->stats.FinishedOps(&db_, db_.db, 1, kSeek);
+      thread->stats.FinishedOps(&SingleDb(), SingleDb().db, 1, kSeek);
```
Yuval-Ariel, line 6706: FinishedOps comment: 
this is will actually fail if the user runs with multi db since SingleDb() asserts we're in a single db state but the rocksdb code simply has a bug here where they just record the stats on the single db. the FinishedOps func can handle &db_ as nullptr. 
im not sure how recording the stats on different dbs when they are in use works so i agree that theres a problem here but lets leave it to another issue and just resort to the vanilla behavior (dbs_to_use[0])

tools/db_bench_tool.cc
```diff
@@ -7261,7 +7302,7 @@ class Benchmark {
         }
         get_weight--;
         gets_done++;
-        thread->stats.FinishedOps(&db_, db_.db, 1, kRead);
+        thread->stats.FinishedOps(&single_db, single_db.db, 1, kRead);
```
Yuval-Ariel, line 7305: same as "FinishedOps comment"

tools/db_bench_tool.cc
```diff
@@ -7271,15 +7312,15 @@ class Benchmark {
         }
         put_weight--;
         puts_done++;
-        thread->stats.FinishedOps(&db_, db_.db, 1, kWrite);
+        thread->stats.FinishedOps(&single_db, single_db.db, 1, kWrite);
```
Yuval-Ariel, line 7315: same as "FinishedOps comment"

tools/db_bench_tool.cc
```diff
@@ -7271,15 +7312,15 @@ class Benchmark {
...
       } else if (delete_weight > 0) {
         Status s = DeleteMany(db, write_options_, key);
         if (!s.ok()) {
           ErrorExit("deletemany error: %s", s.ToString().c_str());
         }
         delete_weight--;
         deletes_done++;
-        thread->stats.FinishedOps(&db_, db_.db, 1, kDelete);
+        thread->stats.FinishedOps(&single_db, single_db.db, 1, kDelete);
```
Yuval-Ariel, line 7323: same as "FinishedOps comment"

tools/db_bench_tool.cc
```diff
@@ -7235,6 +7274,8 @@ class Benchmark {
     int64_t puts_done = 0;
     int64_t deletes_done = 0;
 
+    auto& single_db = SingleDb();
```
Yuval-Ariel, line 7277: the RandomWithVerify test is supported for multi db as well. it just uses the db_ to record the metrics but in this case they chose the db with `DB* db = SelectDB(thread);` so lets just use this db?

tools/db_bench_tool.cc
```diff
@@ -7490,7 +7531,7 @@ class Benchmark {
               RateLimiter::OpType::kRead);
         }
 
-        thread->stats.FinishedOps(&db_, db_.db, 1, kSeek);
+        thread->stats.FinishedOps(&SingleDb(), SingleDb().db, 1, kSeek);
```
Yuval-Ariel, line 7534: same as "FinishedOps comment"

tools/db_bench_tool.cc
```diff
@@ -8002,6 +8043,8 @@ class Benchmark {
   // RandomTransactionVerify() will then validate the correctness of the results
   // by checking if the sum of all keys in each set is the same.
   void RandomTransaction(ThreadState* thread) {
+    auto& single_db = SingleDb();
```
Yuval-Ariel, line 8046: can you move it just above the while loop? so its closer to its use? the verification that its not a multi db is done anyways in Line 8064.

tools/db_bench_tool.cc
```diff
@@ -8295,7 +8338,10 @@ class Benchmark {
         ErrorExit("put error: %s", s.ToString().c_str());
       }
       bytes = key.size() + val.size();
-      thread->stats.FinishedOps(&db_, db_.db, 1, kWrite);
+      // TODO - If there is a single db => no point selecting one above.
+      // If there are multiple db-s, db_ / SingleDb() would be null / fail
+      // => Seems like a bug or suitable only for the single db mode
+      thread->stats.FinishedOps(&SingleDb(), SingleDb().db, 1, kWrite);
```
Yuval-Ariel, line 8344: same as "FinishedOps comment"

tools/db_bench_tool.cc
Yuval-Ariel: part of the RandomWithVerify test

tools/db_bench_tool.cc
Yuval-Ariel: part of the RandomWithVerify test

tools/db_bench_tool.cc
Yuval-Ariel: part of the RandomWithVerify test

tools/db_bench_tool.cc
```diff
@@ -447,6 +449,11 @@ DEFINE_int32(user_timestamp_size, 0,
 DEFINE_int32(num_multi_db, 0,
              "Number of DBs used in the benchmark. 0 means single DB.");
 
+DEFINE_string(
+    dbs_to_use, "",
+    "Indices of the DBs to actually use in the benchmark of all available DBs "
+    " \"\" means use all available DBs. Indices may be specified in any order");
```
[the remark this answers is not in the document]
udi-speedb: Done

tools/db_bench_tool.cc
```diff
@@ -2639,6 +2651,71 @@ class Benchmark {
...
+  // Use this to access the DB when context requires a Single-DB mode
+  DBWithColumnFamilies& SingleDb() {
+    if (IsSingleDb() == false) {
+      ErrorExit("Expecting a Single DB but thare are %" PRIu64 " DB-s",
+                NumDbs());
+    }
+    return dbs_[0];
+  }
+
+  DBWithColumnFamilies& FirstDb() { return dbs_[0]; }
```
Yuval-Ariel, line 2663: shouldnt this be `dbs_to_use_[0]` ? the first db from the current active ones in the group
same for line 2660 {SingleDb()}

tools/db_bench_tool.cc
```diff
@@ -3656,7 +3728,8 @@ class Benchmark {
             ErrorExit("Encountered an error starting a trace, %s",
                       s.ToString().c_str());
           }
-          s = db_.db->StartTrace(trace_options_, std::move(trace_writer));
+          s = SingleDb().db->StartTrace(trace_options_,
```
Yuval-Ariel, line 3731: should this and the following SingleDb() call also be replaced with FirstDb()?, these are options that we're not currently using (the trace na block cache trace) but i dont see that these are only valid for single db mode

Yuval-Ariel: `SingleDb().db->EndTrace()`
`SingleDb().db->EndBlockCacheTrace()` as well

udi-speedb: Currently, in a multi-db scenario, db_.db is nullptr.
So, the code db_.db->XXXX() will crash.
Replacing db_.db->XXX with SingleDB()->XXX() will at least abort gracefully with a reasonable error message which is an improvement.

Yuval-Ariel: fair enough

Yuval-Ariel (approved):

Issue #277, opened 2022-11-28
db_bench: bug ErrorExit called on static function


Yuval-Ariel:
ErrorExit from the Benchmark namespace was called on a static function which led to the error :
`tools/db_bench_tool.cc:3276:67: error: cannot call member function ‘void rocksdb::Benchmark::ErrorExit(const char*, ...)’ without object
 3276 |       ErrorExit("Old clock cache implementation has been removed.");`

fix: call the outer scope ErrorExit function.

also, remove the need to add an exit(1) after ErrorExit to stop compiler complaining about : 
`error: control reaches end of non-void function`

PR #278, opened 2022-11-28
db_bench: ErrorExit from static func bug (#277)

Yuval-Ariel:
while we're at it, also, remove the need to add an exit(1) after ErrorExit to stop compiler complaining about :
error: control reaches end of non-void function

### DIFF
--- a/tools/db_bench_tool.cc
+++ b/tools/db_bench_tool.cc
@@ -127,7 +127,7 @@ namespace {
 // between groups, and destroyed after running the last group
 std::unique_ptr<ROCKSDB_NAMESPACE::Benchmark> benchmark;
 
-void ErrorExit(const char* format, ...) {
+int ErrorExit(const char* format, ...) {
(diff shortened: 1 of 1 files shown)


tools/db_bench_tool.cc
```diff
@@ -4738,11 +4732,9 @@ class Benchmark {
 
             options.blob_cache = NewLRUCache(co);
           } else {
-            fprintf(
-                stderr,
+            exit(ErrorExit(
```
udi-speedb, line 4735: This is in a void method so I think no need to exit(ErrorExit( just plain ErrorExit would do.

udi-speedb:
I suggest replacing the word "bug" with "compilation error" as the code didn't compile rather than being a bug

tools/db_bench_tool.cc
Yuval-Ariel: done

PR #294, opened 2022-12-13
Add RefreshOptions method to support live config changes

mrambacher:
Added a method and options to support reading configuration changes to options periodically while a process is running.  Every "refresh_options_sec", the periodic task scheduler runs a task to check if the "refresh_options_file" exists.  If it does, the file is loaded as an options file and the mutable options from the file are updated in the running process (via SetDBOptions and SetOptions).  Once updated, the refresh file is deleted.  If an error occurs, it will be written to the log file.

Note that the file is read as an Options file and must be a valid Options file.  This means that the file must have a DBOptions section and at least one ColumnFamilyOptions section.  Only options that are mutable can be in this file -- immutable options will cause an error and abort the processing.

Additionally, note that TableFactory options must be specified as part of the ColumnFamilyOptions.  For example, to change the block size, the "table_factory.block_size" option should be set in the appropriate ColumnFamilyOptions.  Options set in a TableFactory section will be ignored.

Currently tested manually via db_bench.  Unit tests will be forthcoming.

### DIFF
--- a/db/db_impl/db_impl.cc
+++ b/db/db_impl/db_impl.cc
@@ -1155,6 +1155,7 @@ void DBImpl::RefreshOptions() {
   }
   TEST_SYNC_POINT("DBImpl::RefreshOptions::Start");
   Status s = fs_->FileExists(new_options_file, IOOptions(), nullptr);
+  TEST_SYNC_POINT_CALLBACK("DBImpl::RefreshOptions::FileExists", &s);
   if (!s.ok()) {
     return;
   }
@@ -1167,12 +1168,14 @@ void DBImpl::RefreshOptions() {
   cfg_opts.mutable_options_only = true;
   RocksDBOptionsParser op;
   s = op.Parse(cfg_opts, new_options_file, fs_.get());
+  TEST_SYNC_POINT_CALLBACK("DBImpl::RefreshOptions::Parse", &s);
   if (!s.ok()) {
     ROCKS_LOG_WARN(immutable_db_options_.info_log,
                    "Failed to parse Options file (%s): %s\n",
                    new_options_file.c_str(), s.ToString().c_str());
   } else if (!op.db_opt_map()->empty()) {
     s = SetDBOptions(*(op.db_opt_map()));
+    TEST_SYNC_POINT_CALLBACK("DBImpl::RefreshOptions::SetDBOptions", &s);
     if (!s.ok()) {
       ROCKS_LOG_WARN(immutable_db_options_.info_log,
                      "Failed to refresh DBOptions, Aborting: %s\n",
@@ -1191,6 +1194,7 @@ void DBImpl::RefreshOptions() {
                          cf_name.c_str());
         } else if (!cfd->IsDropped()) {
           s = SetCFOptionsImpl(cfd, cf_opt_map);
+          TEST_SYNC_POINT_CALLBACK("DBImpl::RefreshOptions::SetCFOptions", &s);
           if (!s.ok()) {
             ROCKS_LOG_WARN(immutable_db_options_.info_log,
                            "Failed to refresh CFOptions for CF %s: %s\n",
@@ -1202,9 +1206,11 @@ void DBImpl::RefreshOptions() {
     }
   }
   s = fs_->DeleteFile(new_options_file, IOOptions(), nullptr);
+  TEST_SYNC_POINT_CALLBACK("DBImpl::RefreshOptions::DeleteFile", &s);
   ROCKS_LOG_INFO(immutable_db_options_.info_log,
                  "RefreshOptions Complete, deleting options file %s: %s\n",
                  new_options_file.c_str(), s.ToString().c_str());
+  TEST_SYNC_POINT("DBImpl::RefreshOptions::Complete");
 }
 #endif  // ROCKSDB_LITE
 
--- a/db/db_impl/db_impl.cc
+++ b/db/db_impl/db_impl.cc
@@ -276,6 +276,8 @@ DBImpl::DBImpl(const DBOptions& options, const std::string& dbname,
   periodic_task_functions_.emplace(
       PeriodicTaskType::kRecordSeqnoTime,
       [this]() { this->RecordSeqnoToTimeMapping(); });
+  periodic_task_functions_.emplace(PeriodicTaskType::kRefreshOptions,
+                                   [this]() { this->RefreshOptions(); });
 #endif  // ROCKSDB_LITE
 
   versions_.reset(new VersionSet(dbname_, &immutable_db_options_, file_options_,
@@ -829,6 +831,15 @@ Status DBImpl::StartPeriodicTaskScheduler() {
       return s;
     }
   }
+  if (mutable_db_options_.refresh_options_sec > 0) {
+    Status s = periodic_task_scheduler_.Register(
+        PeriodicTaskType::kRefreshOptions,
+        periodic_task_functions_.at(PeriodicTaskType::kRefreshOptions),
+        mutable_db_options_.refresh_options_sec);
+    if (!s.ok()) {
+      return s;
+    }
+  }
 
   Status s = periodic_task_scheduler_.Register(
       PeriodicTaskType::kFlushInfoLog,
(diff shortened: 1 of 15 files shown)


udi-speedb:
@mrambacher - Is there an issue for this PR?
If not, please create one.
Either way, please link the issue with this PR. Thanks

udi-speedb:
Why is this feature not available in RocksDB Lite? 
Actually, for my edification, what should be the capabilities / restrictions on RocksDB lite in general? 
Are you aware of any resource specifying this?

udi-speedb:
I think we should add support for this feature in db_stress and crash testing

db/db_impl/db_impl.cc
```diff
@@ -1127,6 +1138,76 @@ void DBImpl::FlushInfoLog() {
...
+      if (!cf_opt_map.empty()) {
+        const auto& cf_name = (*op.cf_names())[idx];
+        auto cfd = versions_->GetColumnFamilySet()->GetColumnFamily(cf_name);
+        if (cfd == nullptr) {
+          ROCKS_LOG_WARN(immutable_db_options_.info_log,
+                         "RefreshOptions failed locating CF: %s\n",
+                         cf_name.c_str());
+        } else if (!cfd->IsDropped()) {
+          s = SetCFOptionsImpl(cfd, cf_opt_map);
+          if (!s.ok()) {
```
udi-speedb, line 1194: Does failure to update a cf's options mean we end up with a partial update?
We may have updated some of the cf-s and we'll skip the rest.

db/db_impl/db_impl.cc
```diff
@@ -1127,6 +1138,76 @@ void DBImpl::FlushInfoLog() {
...
+            ROCKS_LOG_WARN(immutable_db_options_.info_log,
+                           "Failed to refresh CFOptions for CF %s: %s\n",
+                           cf_name.c_str(), s.ToString().c_str());
+          }
+        }
+      }
+      idx++;
+    }
+  }
+  s = fs_->DeleteFile(new_options_file, IOOptions(), nullptr);
```
udi-speedb, line 1204: In case we have failed deleting this file (which may or may not be a valid options file that we have just successfully updated or failed to update), does this mean we will repeat the process as long as the file is there?

db/db_impl/db_impl.cc
```diff
@@ -1127,6 +1138,76 @@ void DBImpl::FlushInfoLog() {
...
+                     "Failed to refresh DBOptions, Aborting: %s\n",
+                     s.ToString().c_str());
+    }
+  }
+  if (s.ok()) {
+    int idx = 0;
+    for (const auto& cf_opt_map : *(op.cf_opt_maps())) {
+      if (!cf_opt_map.empty()) {
+        const auto& cf_name = (*op.cf_names())[idx];
+        auto cfd = versions_->GetColumnFamilySet()->GetColumnFamily(cf_name);
```
udi-speedb, line 1187: Is it possible to hold a pointer a dropped CF at this point?
If it is possible, is there any need or benefit in avoiding the update to a dropped CF?

udi-speedb (changes requested):

mrambacher:
> Why is this feature not available in RocksDB Lite? Actually, for my edification, what should be the capabilities / restrictions on RocksDB lite in general? Are you aware of any resource specifying this?

The PeriodicTaskScheduler is not available in LITE mode.  I do not know why.  I do not know how things were decided LITE or not.

db/db_impl/db_impl.cc
```diff
@@ -1127,6 +1138,76 @@ void DBImpl::FlushInfoLog() {
   LogFlush(immutable_db_options_.info_log);
 }
 
+#ifndef ROCKSDB_LITE
+// Periodically checks to see if the new options should be loaded into the
+// process. log.
+void DBImpl::RefreshOptions() {
+  if (shutdown_initiated_) {
+    return;
+  }
```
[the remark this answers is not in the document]
mrambacher: This is exactly the same code as under the other PeriodicTasks (like RefreshStats).  I am not positive of the rules for the mutex_, but followed the other examples.

options/db_options.cc
```diff
@@ -1002,6 +1011,8 @@ MutableDBOptions::MutableDBOptions(const DBOptions& options)
           options.delete_obsolete_files_period_micros),
       stats_dump_period_sec(options.stats_dump_period_sec),
       stats_persist_period_sec(options.stats_persist_period_sec),
+      refresh_options_sec(options.refresh_options_sec),
+      refresh_options_file(options.refresh_options_file),
```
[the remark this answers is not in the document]
udi-speedb: Oh, I haven't noticed that it's not a standard copy ctor.

db/db_impl/db_impl.cc
```diff
@@ -829,6 +831,15 @@ Status DBImpl::StartPeriodicTaskScheduler() {
       return s;
     }
   }
+  if (mutable_db_options_.refresh_options_sec > 0) {
```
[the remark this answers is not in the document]
mrambacher: If the file is not set, the code defaults to "Options.new"

db/db_impl/db_impl.cc
mrambacher: If updating the DBOptions fails, the entire process fails.

If any of the ColumnFamilyOptions updates fails, the DBOptions will be updated (if changed) and any other ColumnFamilyOptions will still be updated.  It is extremely difficult to roll back to the previous version of the options, as you cannot tell if any background task (such as compaction) started during the refresh options process.  Because of this, the code prints an error if the options upgrade fails but continues anyway.

db/db_options_test.cc
```diff
@@ -1152,6 +1154,185 @@ TEST_F(DBOptionsTest, ChangeCompression) {
...
+  ConfigOptions config_options;
+
+  // Test setting an immutable DBOption and see the value
+  // did not change
+  std::unique_ptr<Env> mock(MockEnv::Create(options.env));
+  options.env = mock.get();
+  ASSERT_OK(PersistRocksDBOptions(config_options, options, {"default"},
+                                  {options}, options.refresh_options_file,
+                                  fs.get()));
+  TEST_SYNC_POINT("DBOptionsTest::WaitForUpdates");
```
udi-speedb, line 1217: I think the following race might happen:
- The DBImpl::RefreshOptions() will run and reach the DBImpl::RefreshOptions::Complete sync point
- PersistRocksDBOptions() called
- DBOptionsTest::WaitForUpdates sync point
In that case the test will pass, but without actually testing what you intend here.

May also apply to other unit tests here.
 I think you should first persist the options file, and only then enable sync point processing. That way you would ensure that the refresh task occurred after the options file you have prepared in on disk.

Another way (and also something you need to test IMO is to verify that the file you have prepared was indeed deleted before checking that the test has performed as expected. That would guarantee that the file was used for a refresh.

db/db_options_test.cc
```diff
@@ -1152,6 +1154,185 @@ TEST_F(DBOptionsTest, ChangeCompression) {
...
+  SyncPoint::GetInstance()->DisableProcessing();
+  SyncPoint::GetInstance()->LoadDependency(
+      {{"DBImpl::RefreshOptions::Complete", "DBOptionsTest::WaitForUpdates"}});
+  SyncPoint::GetInstance()->EnableProcessing();
+
+  // Test with a file that contains no DBOptions section
+  ASSERT_OK(CreateFile(fs, options.refresh_options_file,
+                       "[CFOptions \"default\"]\n", false));
+  TEST_SYNC_POINT("DBOptionsTest::WaitForUpdates");
```
udi-speedb, line 1270: I would prefer to see that a file that contains just an empty DBOptions + CFOptions passes parsing and then test files that contains just one of them.

db/db_options_test.cc
```diff
@@ -1152,6 +1154,185 @@ TEST_F(DBOptionsTest, ChangeCompression) {
...
+  SyncPoint::GetInstance()->LoadDependency(
+      {{"DBImpl::RefreshOptions::Complete", "DBOptionsTest::WaitForUpdates"}});
+  SyncPoint::GetInstance()->SetCallBack("DBImpl::RefreshOptions::Parse",
+                                        [&](void* arg) {
+                                          auto s = static_cast<Status*>(arg);
+                                          ASSERT_NOK(*s);
+                                        });
+  SyncPoint::GetInstance()->EnableProcessing();
+  // Test with a file that is not an options file
+  ASSERT_OK(CreateFile(fs, options.refresh_options_file, "fred", false));
```
udi-speedb, line 1260: It could have failed at this point already by looking for Options.New and just not finding it. You have no verification that the callback was indeed called.

db/db_options_test.cc
```diff
@@ -1152,6 +1154,185 @@ TEST_F(DBOptionsTest, ChangeCompression) {
...
+  ASSERT_OK(CreateFile(fs, options.refresh_options_file,
+                       "[DBOptions]\n"
+                       "[CFOptions \"default\"]\n"
+                       "unknown = value\n",
+                       false));
+  TEST_SYNC_POINT("DBOptionsTest::WaitForUpdates");
+
+  SyncPoint::GetInstance()->DisableProcessing();
+}
+
```
udi-speedb, line 1335: A few comments about the unit tests:
- You are using the name Options.New as the file name in the test while the default is Options.new. It might be preferable to pick a name that will be different also in case insensitive file systems (Windows?).
- I suggest adding tests for the following cases:
  - That the file is indeed deleted after a refresh cycle?
  - The specified file in not found during a refresh cycle (refresh_options_file is not empty)
  - refresh_options_file is empty but the file by the default name Options.new is found as expected and updates the options
  - The file specified is a binary file
  - Multiple refreshes, each time with a different contents of the same file
  - Multiple refreshes, the name of the file changes 
  - Having options for a CF that doesn't exist 
  - An invalid file name (invalid characters in the path)
-

udi-speedb (changes requested):

db_stress_tool/db_stress_test_base.cc
```diff
@@ -3136,6 +3136,10 @@ void InitializeOptionsFromFlags(
       FLAGS_verify_sst_unique_id_in_manifest;
   options.memtable_protection_bytes_per_key =
       FLAGS_memtable_protection_bytes_per_key;
+#ifndef ROCKSDB_LITE
+  options.refresh_options_sec = FLAGS_refresh_options_sec;
+  options.refresh_options_file = FLAGS_refresh_options_file;
+#endif  // ROCKSDB_LITE
```
udi-speedb, line 3142: This will allow the db_stress user to set a refresh rate & file. However, to actually stress test this feature we need to be able to actually create options files to be used by the production code.

udi-speedb (changes requested):

db/db_options_test.cc
```diff
@@ -1152,6 +1154,185 @@ TEST_F(DBOptionsTest, ChangeCompression) {
...
+  DBOptions new_db_opts = db_->GetDBOptions();
+  ASSERT_EQ(new_db_opts.max_background_jobs, 10);
+  ASSERT_EQ(new_db_opts.max_background_compactions, 20);
+  auto dcfh = db_->DefaultColumnFamily();
+  ColumnFamilyDescriptor dcd;
+  ASSERT_OK(dcfh->GetDescriptor(&dcd));
+  ASSERT_EQ(dcd.options.periodic_compaction_seconds, 200);
+
+  SyncPoint::GetInstance()->DisableProcessing();
```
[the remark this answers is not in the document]
mrambacher: We can/could/should.  I would consider this a general issue that should be solved in the base test class destructor myself and handled generally.  

If you wish for me to fix it for this specific test, I can but wonder about scope/feature creep.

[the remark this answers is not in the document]
udi-speedb: I agree. However, adding new tests that duplicate the cleanup code is bad practice IMO. I would just add a TearDown() to DBOptionsTest to do the cleanup and remove it from your **new** unit tests.

db_stress_tool/db_stress_test_base.cc
mrambacher: I agree that we need such a tool.  I hope to work on one in the near future under a separate PR (for general use as well as this stress test).

db/db_options_test.cc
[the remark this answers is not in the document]
mrambacher: Added code to the destructor.

db/db_options_test.cc
mrambacher: Added a method in the test to deal with the sync points and validate that the file is deleted upon return.  When the file is valid, the test also will validate that the settings have been applied.

db/db_options_test.cc
mrambacher: Added a method to validate that the options file that was written is deleted after the sync point.

udi-speedb (approved):

Yuval-Ariel:
@mrambacher  , it seems the checks have failed.. how is this in QA?

udi-speedb:
@Yuval-Ariel I have already reviewed this. Has anything changed that needs another review?

Yuval-Ariel:
@udi-speedb all the approved reviews were removed for some reason.. and the merging needs at least one approval.

udi-speedb:
Oh, I see. OK. Evil @maxb-io modified (per my request) the policy so that whenever a push is made to an already approved PR, the approval is revoked and the PR needs to be approved again. This is to verify that if anyone updates an already approved PR (usually just squashing / rebasing), it will not go unnoticed and retain its approved state in case an (not less evil) person pushed real content.

udi-speedb (approved):

udi-speedb (approved):

PR #308, opened 2022-12-21
Options addr offset

mrambacher:
This is a merge of PR#[10479](https://github.com/facebook/rocksdb/pull/10479) from RocksDB into Speedb.

When an OptionsTypeInfo is defined, it is passed the offset of the field in question.  This offset is the relative location of the field in the structure being analyzed.  This PR adds a "useBaseAddress" flag that tells the option system to use the base address rather than the offset to any custom functions.

This new flag is useful when there are relationships between fields in the structure.  For example:
- If a std::shared_ptr<Env> shared_env value is added to the DBOptions, the "shared" and "env" can be marked as useBase.  This would allow the system to serialize either of the values (either shared or env and not both) and allow the Create method to update the other value.
- For things like Compaction, the different Options (CompactionOptionsFIFO/Universal) could be skipped during serialization if they did not apply (so if FIFO was specified, do not store or validate Universal options).
- Fields that are specific to certain implementations could be moved/associated with the implementation.  
- When multiple fields are validated/prepared/sanitized together, the code can be placed in the OptionsTypeInfo rather than writing class-specific implementations of these methods.  For example, the BlockBasedTableOptions have fields that are related.  Using this new flag allows the relationship to be initialized and validated without the class methods.


include/rocksdb/utilities/options_type.h
```diff
@@ -925,6 +942,22 @@ class OptionTypeInfo {
   constexpr static const char* kIdPropName() { return "id"; }
   constexpr static const char* kIdPropSuffix() { return ".id"; }
 
+  // Fix user-supplied options to be reasonable
```
RoyBenMoshe, line 945: // There is already a ClipToRange() utility in db_impl.h - Why not add this
// utility to some utils file and use everywhere (avoid code duplication)?

options/options_helper.cc
```diff
@@ -928,7 +928,9 @@ Status OptionTypeInfo::Parse(const ConfigOptions& config_options,
     } else if (parse_func_ != nullptr) {
       ConfigOptions copy = config_options;
       copy.invoke_prepare_options = false;
-      void* opt_addr = GetOffset(opt_ptr);
+      auto opt_addr = IsEnabled(OptionTypeFlags::kUseBaseAddress)
```
RoyBenMoshe, line 931: maybe it is possible to wrap this code in a utility function and call it rather than duplicate the code repeatedly

options/options_helper.cc
```diff
@@ -928,7 +928,9 @@ Status OptionTypeInfo::Parse(const ConfigOptions& config_options,
...
+                          ? opt_ptr
+                          : GetOffset(opt_ptr);
       return parse_func_(copy, opt_name, opt_value, opt_addr);
     } else if (ParseOptionHelper(GetOffset(opt_ptr), type_, opt_value)) {
```
RoyBenMoshe, line 935: Is there an assumption that if the parse_func is null kUseBaseAddress
is disabled?
Same question for any other place using GetOffset() - Are
all of these uses OK? Is it not possible to have a kUseBaseAddress in
these cases?

table/block_based/block_based_table_factory.cc
```diff
@@ -280,36 +277,166 @@ static std::unordered_map<std::string, OptionTypeInfo>
...
         {"no_block_cache",
-         {offsetof(struct BlockBasedTableOptions, no_block_cache),
-          OptionType::kBoolean, OptionVerificationType::kNormal,
-          OptionTypeFlags::kNone}},
+         OptionTypeInfo(offsetof(struct BlockBasedTableOptions, no_block_cache),
+                        OptionType::kBoolean, OptionVerificationType::kNormal,
+                        OptionTypeFlags::kUseBaseAddress)
+             .SetPrepareFunc([](const ConfigOptions& /*opts*/,
+                                const std::string& /*name*/, void* addr) {
+               auto bbto = static_cast<BlockBasedTableOptions*>(addr);
```
RoyBenMoshe, line 285: However, if this option is handled before the block_cache
option, then a block cache will be created.
Does this depend in any way on the order?
There is a block_cache option that is handled separately What is this
option is handled first, sees that bbto->block_cache is null
and creates a block cache. It looks harmless,
the cache created here would be replaced by the actual cache
that would be created there

This code is copied from BlockBasedTableFactory::BlockBasedTableFactory() - Please avoid the duplication The following line wasn't copied - is
that intentional?
co.low_pri_pool_ratio = 0.0;
Anyway, lets try to avoid copying

table/block_based/block_based_table_factory.cc
```diff
@@ -280,36 +277,166 @@ static std::unordered_map<std::string, OptionTypeInfo>
...
+               *interval = ParseInt(value);
+               *interval = std::max(1, *interval);
+               return Status::OK();
+             })
+             .SetPrepareFunc([](const ConfigOptions& /*opts*/,
+                                const std::string& /*name*/, void* addr) {
+               auto interval = static_cast<int*>(addr);
+               *interval = std::max(1, *interval);
+               return Status::OK();
+             })},
```
RoyBenMoshe, line 343: How is it handled at the moment without this capability?
Couldn't we implement the  SetParseFunc() before? Or is it now
mandatory because we are using kUseBaseAddress?
Duplicates (to a degree) in the index_type option

table/block_based/block_based_table_factory.cc
```diff
@@ -280,36 +277,166 @@ static std::unordered_map<std::string, OptionTypeInfo>
...
+               }
+               if (bbto->partition_filters &&
+                   bbto->index_type !=
+                       BlockBasedTableOptions::kTwoLevelIndexSearch) {
+                 // We do not support partitioned filters without partitioning
+                 // indexes
+                 bbto->partition_filters = false;
+               }
+               return Status::OK();
+             })},
```
RoyBenMoshe, line 415: Again, Is  that depends on the order and the
implications? And what happens currently without the feature?

include/rocksdb/utilities/options_type.h
```diff
@@ -101,14 +101,16 @@ enum class OptionTypeFlags : uint32_t {
   kCompareLoose = ConfigOptions::kSanityLevelLooselyCompatible,
   kCompareExact = ConfigOptions::kSanityLevelExactMatch,
```
RoyBenMoshe, line 103: Please open an issue in speedb for the PR and add the issue number to the commit message.

options/options_helper.cc
mrambacher: The kUseBaseAddress is only used if the function is non-null.  This is true for all of the methods (parse, equals, serialize, prepare, etc).  This allows the developer to override the behavior for specific methods (such as parse) but leave it as the default for others (such as serialize).

include/rocksdb/utilities/options_type.h
mrambacher: Removed db_impl ClipToRange

options/options_helper.cc
mrambacher: Added GetBaseOffset function.

table/block_based/block_based_table_factory.cc
mrambacher: I do not believe there will be a race here.  This code is done on initializing the options, which is done after they are all read from the map.  If block_cache is processed first, it means that the cache will be prepared unnecessarily before it is cleared/reset by this method.

table/block_based/block_based_table_factory.cc
mrambacher: I do not believe there is any order dependency here because these are "initialization" done after the fields are already set.

mrambacher:
Experiencing issues with this git branch.  Abandoning this PR in favor of #397.

PR #311, opened 2022-12-23
Delayed Writes: fix L0 calc bug

Yuval-Ariel:


db/column_family.cc
```diff
@@ -1013,15 +1013,15 @@ ColumnFamilyData::CalculateWriteDelayDividerAndMaybeUpdateWriteStallCause(
 
   // L0 files
   double l0_divider = 1;
-  const auto extra_l0_ssts = vstorage->NumLevelFiles(0) -
+  const auto extra_l0_ssts = vstorage->l0_delay_trigger_count() -
```
ayulas, line 1016: it means that you used the incorrect value to calc the write delay , right?

ayulas (approved):

db/column_family.cc
Yuval-Ariel: i think that l0_delay_trigger_count()  is used generically which works for universal compaction as well. it worked just as well before but i noticed that l0_delay_trigger_count()  is used to count the L0 files before and after that point in the code..

Issue #312, opened 2022-12-25
WBM: Print to the log info about WBM's cost to cache


udi-speedb:
A user creating a WBM may or may not specify a cache to use for costing the allocations reported to the WBM to the specified cache.
It will be useful to display this information as well in the log.

PR #313, opened 2022-12-25
Print WBM's cache info to log (#312)

udi-speedb:


memtable/write_buffer_manager.cc
```diff
@@ -285,6 +285,16 @@ std::string WriteBufferManager::GetPrintableOptions() const {
...
   ret.append(buffer);
 
+  const Cache* cache = nullptr;
+  if (cache_res_mgr_ != nullptr) {
+    cache =
+        static_cast<CacheReservationManagerImpl<CacheEntryRole::kWriteBuffer>*>(
+            cache_res_mgr_.get())
+            ->TEST_GetCache();
+  }
+  snprintf(buffer, kBufferSize, "%*s: %p\n", field_width, "wbm.cache", cache);
```
Yuval-Ariel, line 295: whats field_width exactly? last time you increased this value..

Yuval-Ariel (approved):

memtable/write_buffer_manager.cc
udi-speedb: It's set so the printing of the wbm fields will be right-adjusted properly, as seen here:

Screenshot from 2022-12-25 16-42-45

Issue #61, opened 2022-07-20
db_bench: many default values are not the default in Speedb


Yuval-Ariel:
these params are:

table_cache_numshardbits is 4 while default is 6. 
new table reader for comp inputs
block cache 8mb
index shortening mode
enable_pipelined_write - true in db bench and false otherwise.
delayed_write_rate = 8Mb and default value is 0.

and possibly more.

fix by making db_bench take the default value

Yuval-Ariel:
changes:
index_shortening_mode - from 2 (kShortenSeparatorsAndSuccessor) to 1 (kShortenSeparators)
hard_pending_compaction_bytes_limit - from 128 Gb to 256 Gb
delayed_write_rate - from 8Mb to 0 (0 is sanitized to 16Mb) 
enable_pipelined_write - from true to false

erez-speedb:
Seems the rocksdb default are better then the db_bench defaults
Small obj
https://admin.speedb.io/performance?items=2R2iIbfzE1yWcnSIkwUQ&items=0YnC1biInuhalItBnuX6&colors=%23F06292&colors=%2300796b
Large obj
https://admin.speedb.io/performance?items=PDScQQlgGl9TOZfbs12T&items=DQbr0ntUjYJDz3RLExzO&colors=%23F06292&colors=%2300796b
Small DB
https://admin.speedb.io/performance?items=GFN4BI9Xr0wuX0MDjMF6&items=svLLRKKkMmxENz01Igpz&colors=%23F06292&colors=%2300897b
In small DB the overwrite of current default is 16% better but is all other configurations its the same. I belive is an outlier.

Yuval-Ariel:
@erez-speedb  , what do you mean outlier? the trend looks completely different
![image](https://user-images.githubusercontent.com/105581454/212128575-1f471144-2dc5-41b5-b7f7-77063de53a5a.png)
can you point to previous outliers that behaved like this? if you run two exact hashes, can there be a chance for such a difference?
why not run it again and verify whether its an outlier or not?

erez-speedb:
Running the small db test again
The results are back to normal
https://admin.speedb.io/performance?items=MTnDOa9oyhGa33CyQgci&items=Swqg6xZSgqnVYc71L1mA&colors=%2300796b&colors=%23ec407a

Passed perf tests.

PR #322, opened 2023-01-01
db_bench: set db_bench defaults to speedbs (#61)

Yuval-Ariel:
actual values that have changed:
1. index_shortening_mode - from 2 (kShortenSeparatorsAndSuccessor) to 1 (kShortenSeparators)
2. hard_pending_compaction_bytes_limit - from 128 Gb to 256 Gb
3. delayed_write_rate - from 8Mb to 0 (0 is sanitized to 16Mb)
4. enable_pipelined_write - from true to false


Yuval-Ariel:
@erez-speedb  plz check

ayulas (approved):
pls note that now the enable_pipelined_write is flase and it might change the base result

Issue #273, opened 2022-11-26
rebase7.7 db_test bug - PurgeInfoLogs


Yuval-Ariel:
[ RUN      ] DBTest.PurgeInfoLogs
db/db_test.cc:2600: Failure
Value of: env_->GetChildren(dbname_, &db_files).IsNotFound()
  Actual: false
Expected: true

Yuval-Ariel:
this seems to be a timing issue since when running this test separately it passes (./db_test --gtest_filter=DBTest.PurgeInfoLogs) and the CI which runs the unit tests also passes.

Yuval-Ariel:
fails when running the unit tests serially with J=1

Yuval-Ariel:
the error happens since there are leftovers from previous unit tests. thats why running the unit test separately passed and also clearing the test folder (/tmp/rocksdbtest-1000) just before this unit test was run also helps. 
this also proves why running in parallel didnt have this issue since each test is run separately. 

the reason why there are leftovers is because each unit test in a suite gets the same name if its run in the same process and by the same thread which is what happens when using J=1. this is done in PerThreadDBPath in test_util/testharness.cc . 

the call to DestroyDB cleans the db itself but doesnt clean other dbs or files from that dir. e.g. if db_log_dir is used which is often the case in this suite then the LOGs will be stored in a dir (/db_log_dir) inside the db path and it will not be deleted once the db is deleted. 

theres also an option to clean out completely the db path in each call to DestroyDB but this seemed a better way.

PR #324, opened 2023-01-02
unit tests: fix db_test PurgeInfoLogs by creating unique db path (#273)

Yuval-Ariel:
create a unique db path for each new db opened in a suite. the PerThreadDBPath creates a db path based on the PID and the thread so that every test in the suite (e.g. db_test) has the exact same db path. this leads to unwanted leftovers which caused an error in PurgeInfoLogs.

### DIFF
--- a/test_util/testharness.cc
+++ b/test_util/testharness.cc
@@ -45,12 +45,19 @@ std::string PerThreadDBPath(std::string dir, std::string name) {
   return dir + "/" + name + "_" + GetPidStr() + "_" + std::to_string(tid);
 }
 
+std::string PerThreadDBPath(std::string dir, std::string name,
+                            int64_t time_now) {
+  return dir + "/" + name + "_" + GetPidStr() + "_" + std::to_string(time_now);
+}
+
 std::string PerThreadDBPath(std::string name) {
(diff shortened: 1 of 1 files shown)


Yuval-Ariel:
currently checking if it hasnt broken any existing unit tests - https://github.com/speedb-io/speedb/actions/runs/3824428690

Yuval-Ariel:
both CI running the units tests and running the unit tests serially (J=1) on an azure instance finished successfully.

mrambacher:
Another solution was introduced in https://github.com/facebook/rocksdb/pull/10000, but never merged.  This solution allows tests to be tracked back to their test name, making it easier to find the data associated with a test.

Yuval-Ariel:
looks great, by name is definitely better. why was it not merged?

mrambacher:
> looks great, by name is definitely better. why was it not merged?

Not sure why it was never merged.  There was no discussion outside of the public comments.  Feel free to merge it (or I can later in the week).

Yuval-Ariel:
we've decided that @mrambacher  will port the PR mentioned which does the same fix but with test name instead of a timestamp. 
that PR should also close this issue #273

PR #353, opened 2023-01-05
Use the Test Name for the dbname when running unit tests

mrambacher:
This PR is based on the following [RocksDB PR](https://github.com/facebook/rocksdb/pull/10000).

The name of the database directory can be based on the name of the test itself.  This allows multiple tests in the same test class to use different database directories rather than all going to the same or similar directories.

This change can also help in debugging and KEEP_DB tests, as the directory after the test run can be spotted by its name (which is based on the name of the test).

### DIFF
--- a/db/db_test_util.cc
+++ b/db/db_test_util.cc
@@ -88,7 +88,7 @@ DBTestBase::DBTestBase(const std::string path, bool env_do_fsync)
   env_->SetBackgroundThreads(1, Env::LOW);
   env_->SetBackgroundThreads(1, Env::HIGH);
   env_->skip_fsync_ = !env_do_fsync;
-  dbname_ = test::PerThreadDBPath(env_, path);
+  dbname_ = test::PerThreadDBPath(env_, test::GetTestNameForDB(path));
   alternative_wal_dir_ = dbname_ + "/wal";
   alternative_db_log_dir_ = dbname_ + "/db_log_dir";
   auto options = CurrentOptions();
--- a/test_util/testharness.cc
+++ b/test_util/testharness.cc
@@ -33,6 +33,30 @@ std::string GetPidStr() { return std::to_string(getpid()); }
   }
 }
 
+// If suggested is empty, the name will be <test case>-<test name>
+// Replaces all of the "/" in the test case/name with "_", so that they will not
+// appear as directories
+std::string GetTestNameForDB(const std::string& suggested) {
+  const testing::TestInfo* const test_info =
+      testing::UnitTest::GetInstance()->current_test_info();
+  std::string test_name = test_info->name();
+  std::string test_case = test_info->test_case_name();
+  auto pos = test_case.find("/");
+  if (pos != test_case.npos && !suggested.empty()) {
+    test_case = suggested;
+  } else {
+    while (pos != test_case.npos) {
+      test_case[pos] = '_';
+      pos = test_case.find("/", pos);
+    }
+  }
+  for (pos = test_name.find("/"); pos != test_name.npos;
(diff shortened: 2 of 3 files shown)


Yuval-Ariel:
closes #273

Yuval-Ariel:
atm, each suite running in the same process creates one dir and each test in it is using the same dir. these changes create dir for each test inside the suite which will create much more leftovers as can be seen in compaction_service_test. ( DestroyDB doesnt clean up the dir created by the remote compaction. )
i suggest we also add a better cleanup for successful tests

mrambacher:
> atm, each suite running in the same process creates one dir and each test in it is using the same dir. these changes create dir for each test inside the suite which will create much more leftovers as can be seen in compaction_service_test. ( DestroyDB doesnt clean up the dir created by the remote compaction. ) i suggest we also add a better cleanup for successful tests

I agree the tests should do a better job of cleaning up after themselves.  This PR is not about that at all however.  It is meant to allow one to determine what test left what around.  It is also ensures that the parallel tests use separate directories and can eliminate some of the other information (pid/tid, etc) that are built into the directory/DB names now.  The goal would be to simplify the naming in a future PR.

Yuval-Ariel:
what i meant is this PR causes a huge increase in the leftovers (undeleted files and dirs in the test dir (rocksdbtest-1000)) which seems like a problem so it might be addressed here or in a subsequent PR but the point is to address it.

Yuval-Ariel (approved):
LGTM, just needs rewording of the commit msg and add the PR number

Yuval-Ariel:
fuzzer error doesnt seem to be related to the changes.

Yuval-Ariel:
@mrambacher , i've checked the size of the leftovers and heres what i found: (using the makefile)

**parallel make check:**
mrambacher:DBGTestName - 39Mb
main - 14Mb

**serial make check** (J=1)
mrambacher:DBGTestName - 400Mb
main - 160Mb

do you have an idea why theres a difference between the size of the leftovers of the serial run vs the parallel? could the parallel version not run all the tests that the serial does?

Yuval-Ariel:
just noticed that @isaac-io  was working on it in isaac/test-cleanup

Yuval-Ariel:
we've agreed to add the test executable name before so that its easy to identify the suite to which an individual test belongs to.
for tests that inherit from DBTestBase, each test will now have the name such as: (e.g. the first test in db_basic_test suite)
db_basic_test-DBBasicTest.OpenWhenOpen_ProcessID_ThreadID

we've also agreed to address the cleanup only if the need arises.

Yuval-Ariel:
theres also a problem with the fuzzer caused by a running the CI on a branch that exists in a different repo. @maxb-io  is on it.

mrambacher:
> we've agreed to add the test executable name before so that its easy to identify the suite to which an individual test belongs to. for tests that inherit from DBTestBase, each test will now have the name such as: (e.g. the first test in db_basic_test suite) db_basic_test-DBBasicTest.OpenWhenOpen_ProcessID_ThreadID
> 
> we've also agreed to address the cleanup only if the need arises.

Adding the file name is more complex than I was hoping and I would like to address that in a subsequent PR where we potentially can also clean up the test name further.

Yuval-Ariel:
sure np

Yuval-Ariel:
just needs to be a single commit with a proper msg

Yuval-Ariel:
@mrambacher , can you fix this?

> just needs to be a single commit with a proper msg

Yuval-Ariel:
on hold until commit msg is fixed

Issue #314, opened 2022-12-25
unit tests: enable CompactionServiceTest.RemoteEventListener


Yuval-Ariel:
CompactionServiceTest.RemoteEventListener was disabled since it failed on several branches and wasnt considered a serious issue. fix the unit test correctly

Yuval-Ariel:
running the test separately passes.
i was able to reproduce the failure by running the test in a loop:
while ./compaction_service_test --gtest_filter=CompactionServiceTest.RemoteEventListener; do sleep 1; done
while the system was busy.

error is:
[ RUN      ] CompactionServiceTest.RemoteEventListener
db/compaction/compaction_service_test.cc:782: Failure
Value of: result.second
  Actual: false
Expected: true
db/compaction/compaction_service_test.cc:788: Failure
Value of: num == 1
  Actual: false
Expected: true
[  FAILED  ] CompactionServiceTest.RemoteEventListener (335 ms)

PR #354, opened 2023-01-06
unit tests: fix CompactionServiceTest.RemoteEventListener (#314)

Yuval-Ariel:
the test assumes that each compaction job_id from the remote compaction is unique but in fact all the job_ids are 1.
the test passed thus far since there was just 1 bg compaction by default. once we increased it to 8, there could be concurrent remote compactions which led to this error.

### DIFF
--- a/db/compaction/compaction_service_test.cc
+++ b/db/compaction/compaction_service_test.cc
@@ -772,7 +772,7 @@ TEST_F(CompactionServiceTest, FallbackLocalManual) {
   VerifyTestData();
 }
 
-TEST_F(CompactionServiceTest, DISABLED_RemoteEventListener) {
+TEST_F(CompactionServiceTest, RemoteEventListener) {
   class RemoteEventListenerTest : public EventListener {
    public:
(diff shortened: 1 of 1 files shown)


db/compaction/compaction_service_test.cc
```diff
@@ -817,7 +817,13 @@ TEST_F(CompactionServiceTest, DISABLED_RemoteEventListener) {
   remote_listeners.emplace_back(listener);
 
   Options options = CurrentOptions();
+  options.max_background_compactions = 1;
   ReopenWithCompactionService(&options);
+  // multiple compactions might notify on OnSubcompactionBegin simultaneously
```
ayulas, line 822: but why not support that? the job id is unique in the same thread, so why not change the set to map and then with the thread id and job id you will have only 1 appearance and you can check also many background compactions jobs ?

ayulas (approved):

db/compaction/compaction_service_test.cc
ayulas: agrred this fix is good enough

PR #355, opened 2023-01-06
325 db bench align the tests we're using

Yuval-Ariel:
fix seekrandomwriterandom, BGWriter, readrandom, seekrandom and readrandomwriterandom.

made sure these tests:
1. use one rand num to get the db and cf and the key which makes it possible to find a key in a cf later on.
2. ability to have a read and write limiter where ever relevant.
3. operations are done on all cfs randomly and not only on the default cf

### DIFF
--- a/tools/db_bench_tool.cc
+++ b/tools/db_bench_tool.cc
@@ -2890,6 +2890,32 @@ class Benchmark {
   std::mutex mutex_;
   bool seek_started_;
 
+  inline void LimitReadOrWriteRate(RateLimiter::OpType op_type,
+                                   ThreadState* thread,
+                                   int64_t bytes_to_request) {
+    RateLimiter* rate_limiter_to_use;
+    switch (op_type) {
+      case RateLimiter::OpType::kRead: {
+        rate_limiter_to_use = thread->shared->read_rate_limiter.get();
+        break;
+      }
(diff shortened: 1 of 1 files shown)


udi-speedb:
@Yuval-Ariel 
Points raised during our conversation. Please consider how to handle these:
- The PR contains multiple unrelated fixes / improvements. Consider splitting to a few well-defined topics (e.g., Use all of the applicable cf-s ..., Support rate limiter..., etc.)
- Function to handle the rate limiting logic, potentially one per read and one per write. To be used in the code you have modified rather than creating more duplicate code
- Rename `db` to `db_with_cfh.db` in calls to `FinishedOps(db_with_cfh, db`
- We are changing the current (FB / RocksDB) behaviour of BGWriter in db_bench when the user uses multiple CF-s. This may be a bug fix in RocksDB, or a change of deliberate intention on the part of RocksDB. Regardless, existing users of db_bench may get different results which they may not expect. So, we must clearly document this (release notes / feature documentation / other). In addition, it may or may not make sense to allow users to set some flag that will keep the current behaviour.

Yuval-Ariel:
@udi-speedb  i've addressed the above comments. plz rereview

udi-speedb (approved):

PR #359, opened 2023-01-12
357 upgrade base rocksdb version to 778

Yuval-Ariel:


mrambacher (approved):
In the future, for this "merge from base", it would be helpful to see in the PR the delta between the new Rocks/Speedb and old Rocks/Speedb versions.  In other words, if you find the differences between Speedb and the version of Rocks it was built on, and compare it to the Speedb code with the updated version of Rocks, the changes should be minimal.

Issue #11, opened 2022-06-14
Tracking issue for better memory manager


hilikspdb:
Need to define a memory manager that will allow the system the work at max performance while limiting the usage of memory

- [ ] #12

PR #361, opened 2023-01-14
WIP: Make Speed MemTable default

mrambacher:
If the Speedb memtable is installed, it will be the default


ayulas:
pls check with this branch ayula/default_spdb_memtable

Guyme:
will be sloved with #11

ayulas:
currently the memtable wont be default since it cost memory and its benefit is with the WF

Issue #358, opened 2023-01-11
Dynamic Delay: CFs shouldnt override each other


Yuval-Ariel:
Currently, in a setting of multiple cfs, the delayed_write_rate_ value is overridden in each call to set_delayed_write_rate. this is a problem that existed also when use_dynamic_delay = false but was much less severe since the new delayed rate is always calced from the old. so in the example below, the 2nd cf would set the rate to about 10 Mb/s instead of 512Mb/s.
e.g. a scenario where there are 2 cfs, L0 slowdown is 20, L0 stop is 36 and delayed_write_rate is set to 512Mb/s. 
cf1 has a serious delay state (35 L0 files) and sets the delayed_write_rate_ to 8Mb/s but right after it sets it, cf2 finishes a flush and sets delayed_write_rate_ to 512Mb/s since it only has 20 L0 files. all further writes will run in speed of 512Mb/s until a new InstallSuperVersion happens which is not what should happen.

what should happen is that system will run at the slowest speed required from all cfs. 

do this by storing the delay requirements of all the CFs in the write controller and only set a new delayed_write_rate_ when its smaller than the min.

Yuval-Ariel:
@erez-speedb  , status?

erez-speedb:
Looking at the multi CF test
https://admin.speedb.io/performance?items=s8zPAKKkA6beO0NwPX5Z&items=hKWqp0Aknv7eR3CDlnnG&colors=%23F06292&colors=%23BA68C8
<img width="1229" alt="Screenshot 2023-02-09 at 11 58 33" src="https://user-images.githubusercontent.com/106964746/217780176-699c0a97-c7a6-4362-acac-b1e857866ca1.png">
There is no obvious improvement is stability of performance.

However, there are no degradation in the rest of the tests 
https://admin.speedb.io/performance?items=1fgedOdqXztBsfrXx7W6&items=VhI3ee0518hcWoQ7nGov&colors=%2300796b&colors=%23BA68C8
https://admin.speedb.io/performance?items=bpNEndqhyW9SLaMmftzM&items=DQbr0ntUjYJDz3RLExzO&colors=%23F06292&colors=%2300796b
https://admin.speedb.io/performance?items=5TW4COdc1Nz81dwYliKx&items=zmqpQ9ogatg10p1lqXkC&colors=%23F06292&colors=%2300796b

@Yuval-Ariel  @Guyme  If this improves correctness then we can merge, performance wise, it dost seems to make a difference

PR #364, opened 2023-01-16
Dynamic Delay: CFs wont override each other (#358)

Yuval-Ariel:
Currently, in a setting of multiple cfs, the delayed_write_rate_ value is overridden by the latest cf that entered a delay state. so if a cf registered a delay state that is less severe than the previous cf, it will override the previous value of delayed_write_rate_ with a higher rate than the system should use.

fix it by storing the delay requirements of all the CFs in the write controller and only set a new delayed_write_rate_ when its the min value from all entries stored.

### DIFF
--- a/db/write_controller.cc
+++ b/db/write_controller.cc
@@ -33,6 +33,46 @@ std::unique_ptr<WriteControllerToken> WriteController::GetDelayToken(
   return std::unique_ptr<WriteControllerToken>(new DelayWriteToken(this));
 }
 
+void WriteController::InsertOrAssignToCfIdAndRateMap(uint32_t cf_id,
+                                                     uint64_t write_rate) {
+  cf_id_to_write_rate_.insert_or_assign(cf_id,
+                                        RateAndWasMin{write_rate, false});
+}
+
+uint64_t WriteController::GetMinRate() {
+  assert(!cf_id_to_write_rate_.empty());
+  uint64_t min_rate = std::numeric_limits<uint64_t>::max();
+  uint32_t min_id = 0;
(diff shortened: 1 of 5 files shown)


db/write_controller.h
```diff
@@ -50,6 +50,32 @@ class WriteController {
   // threads will be increased
   std::unique_ptr<WriteControllerToken> GetCompactionPressureToken();
 
+  struct RateAndWasMin {
```
ayulas, line 53: why not using set that the cf pointer is the element (it consists the write rate and id ) and the sorted set function is the write rate so the min is the first element. that way when you when you remove / change/add the you just need call set with the first set element write rate...
you dont need the is min etc ... you dont need all above functions...

db/write_controller.cc
```diff
@@ -33,6 +33,50 @@ std::unique_ptr<WriteControllerToken> WriteController::GetDelayToken(
   return std::unique_ptr<WriteControllerToken>(new DelayWriteToken(this));
 }
 
+void WriteController::InsertOrAssignToCfIdAndRateMap(uint32_t cf_id,
```
ayulas, line 36: see my buttom comments. if you use std set with cf object pointer you dont need this

ayulas (dismissed):
we discussed , and for now its good enough . for future we might think of map  with custom compare

Yuval-Ariel:
i've considered replacing the logic for finding the min rate inside GetMinRate() with std::min_element but i'm setting is_min = false on all elements through iteration anyways..

Yuval-Ariel:
reabased on 2.2.1 since there were major db bench changes then and the recent baseline in on 2.2.1. @ayulas , plx reapprove

PR #366, opened 2023-01-16
Add an immutable option to use clean single delete during flush

ayulas:


hilikspdb:
my review  is only on adding an option to control the feature . this review passed sucessfuly

Guyme (approved):

Yuval-Ariel:
@ayulas , this needs a make format.
also, wheres the actual logic of this option?

ayulas:
new pull request will created with different branch since the branch cannot be changed

Issue #338, opened 2022-08-09
Add index size report per column family


bosmatt:
Customer ask

In order to get better understanding of the current resource utilization, the ask is to report the index size per CF to the log.

udi-speedb:
Notes for performance / QA:
The change consists of counting per-cf sizes when collecting block cache statistics.
There is now, per CF, another line in the log when displaying the block cache statistics. The log line displays, per cf, the sizes in the cache of the index blocks, filter blocks, and data blocks.
So, QA should verify that the data displayed in the log is correct (the sum of all sizes of all the cf-s should be the same as the total for the entire block cache).
Performance-wise, I expect that there will not be any noticeable performance degradation.

Yuval-Ariel:
@erez-speedb  plz check

erez-speedb:
I'm running our multi CF test, and will run the small obj test with 2 CF as well

erez-speedb:
@udi-speedb  Can you pls rebase on latest main?
I see some degradation on the small obj test and I'm not sure if it from the test, the code or the main it was based on.

udi-speedb:
@erez-speedb I have rebased my branch on latest origin/main, built, successfully ran "make check" and pushed (forced) the branch.

erez-speedb:
After the second test, there is a degradation with seeks.
https://admin.speedb.io/performance?items=kXAjadycwIiq5RlgGBZp&items=iJeTPOC4ExuK3i0s3RcJ&items=cbAsF4BziNIhcRaxfO4f&colors=%23F06292&colors=%23BA68C8&colors=%2300796b

PR #368, opened 2023-01-17
Report to Log the Index size per CF - LRU Cache Only (#338)

udi-speedb:


cache/sharded_cache.h
```diff
@@ -82,12 +85,13 @@ class ShardedCache : public Cache {
...
-                        const CacheItemHelper* helper, size_t charge,
-                        Handle** handle = nullptr,
-                        Priority priority = Priority::LOW) override;
+                        DeleterFn deleter, Handle** handle, Priority priority,
+                        Cache::ItemOwnerId item_owner_id) override;
+  virtual Status Insert(
+      const Slice& key, void* value, const CacheItemHelper* helper,
+      size_t charge, Handle** handle = nullptr,
+      Priority priority = Priority::LOW,
+      Cache::ItemOwnerId item_owner_id = kUnknownItemId) override;
```
speedbmike, line 94: Is default value needed here?

cache/cache.cc
```diff
@@ -126,4 +126,74 @@ Status Cache::CreateFromString(const ConfigOptions& config_options,
...
+    allocated_id = next_item_owner_id_++;
+
+    if (allocated_id == kMaxOwnerItemId) {
+      has_wrapped_around_ = true;
+    }
+  } else {
+    std::lock_guard<std::mutex> lock(free_ids_mutex_);
+    // There must be at least 1 list element on the free_ids_ list since
+    // we num_free_ids_ must have been >= 1 on entry if we got here.
+    // Asserting to catch the bug during testing, however, this is a
```
speedbmike, line 164: All that code is to avoid locking mutex when checking if free list is non-empty upfront? But isn't allocation itself very rare? And what happens when another thread changes these values at the same time?

cache/lru_cache.cc
```diff
@@ -468,7 +469,7 @@ void LRUCacheShard::Promote(LRUHandle* e) {
           e->IsHighPri() ? Cache::Priority::HIGH : Cache::Priority::LOW;
       s = Insert(e->key(), e->hash, /*value=*/nullptr, 0,
                  /*deleter=*/nullptr, /*helper=*/nullptr, /*handle=*/nullptr,
-                 priority);
+                 priority, Cache::kUnknownItemId);
```
speedbmike, line 472: Why is it unknown here?

db/table_cache.cc
```diff
@@ -162,15 +162,17 @@ Status TableCache::GetTableReader(
...
+        level, block_cache_tracer_, max_file_size_for_l0_meta_pin,
+        db_session_id_, file_meta.fd.GetNumber(), expected_unique_id,
+        file_meta.fd.largest_seqno);
+    table_reader_options.cache_owner_id = cache_owner_id_;
+
     s = ioptions_.table_factory->NewTableReader(
-        ro,
-        TableReaderOptions(ioptions_, prefix_extractor, file_options,
-                           internal_comparator, skip_filters, immortal_tables_,
-                           false /* force_direct_prefetch */, level,
```
speedbmike, line 169: Just refactoring or any functional change here?

speedbmike (commented):
Many small changes that will make future rebases more difficult. Need to make sure that when HyperClockCache is added instead of ClockCache this functionality won't be left out

db/table_cache.cc
udi-speedb: I need to set the cache_owner_id_ in the TableReaderOptions object. I didn't want to change the ctor of that class and add the cache_owner_id as an additional parameter (optional, with a default value?). Therefore, I created the TableReaderOptions object separately, set the cache_owner_id explicitly, and passed that to the NewTableReader() call.

cache/sharded_cache.h
udi-speedb: > Is default value needed here?
This function overrides the one in Cache, which does have a default value. If I want to allow the user to call Insert() without explicitly specifying the item_owner_id, it must have a default (and It should be the same default). The fact that the base class has a default, doesn't make this inherit that behaviour.

cache/cache.cc
udi-speedb: Yes, the code if for that. The code is supposed to be thread safe (under the assumptions described in the comments). Please let me know if you do find a bug in it.
However, if you feel I should just use the mutex at the start since this is a rare event, that will simplify the code and the code may be removed.

cache/sharded_cache.h
udi-speedb: As for your comment about many small changes. I agree. However, since I am changing the signature of the Insert function and the signature of the callback function on which the solution is based, I see no other option.
If, in addition, I will not use a default value for the item_owner_id, there will be even more changes in calls that currently may stay as is because that default saves us the need to change them as well.

cache/lru_cache.cc
udi-speedb: Agree. I have changed this to pass the item owner id of the promoted handle (assuming that was what bothered you)

speedbmike (approved):

udi-speedb:
@erez-speedb, @Yuval-Ariel I have just noticed that I haven't squashed the commits.
I will now rebase on latest origin/main, squash, and (force) push

udi-speedb:
@speedbmike - I apologize. I did a test on this PR together with @maxb-io. @maxb-io changed the policy to invalidate a pr approval once new commits are pushed to the pr. So I pushed a dummy commit and now I have forced pushed the branch again without the dummy commit.
However, I think you are supposed to re-approve it

Yuval-Ariel (approved):

PR #392, opened 2023-02-07
346 global delayed write rate

### LINKED ISSUE #346: Global delayed write rate
What:
Delayed write rate in a system level - shared to all DBs
Why:
The current delayed write mechanism allows the user to set a limit per database. 
Databases that have internal delay may be starved by others that work faster
Therefore a write-control for the entire system is requires. 
Who:
 -This will mostly relevant for systems with multiple DBs. 

Requirements: 
1. The user shall have the ability to set delayed write rate to each DB if wishes or a global rate for all the DBs.
2. At this stage, the triggers to activate the delayed writes should remain the same. 
3. The system should be able to control the writes rate for the entire system
4. As a result, starvation should be prevented and the system write rate can be predicted.

Yuval-Ariel:
Write Controller is now a shared_ptr option in ImmutableDBOptions which keeps track and enforces all the delay requirements of the cfs in the dbs that those options were passed to. 
closes #346

### DIFF
--- a/db/column_family.cc
+++ b/db/column_family.cc
@@ -760,6 +760,9 @@ void ColumnFamilyData::SetDropped() {
   // can't drop default CF
   assert(id_ != 0);
   dropped_ = true;
+  if (column_family_set_->write_controller()->is_dynamic_delay()) {
+    column_family_set_->DeleteSelfFromMapAndMaybeUpdateDelayRate(id_);
+  }
   write_controller_token_.reset();
 
   // remove from column_family_set
@@ -907,6 +910,10 @@ double ColumnFamilyData::TEST_CalculateWriteDelayDivider(
       compaction_needed_bytes, mutable_cf_options, write_stall_cause);
 }
 
+uint64_t ColumnFamilyData::UpdateCFRate(uint32_t id, uint64_t write_rate) {
+  return column_family_set_->UpdateCFRate(id, write_rate);
+}
+
 std::unique_ptr<WriteControllerToken> ColumnFamilyData::DynamicSetupDelay(
     WriteController* write_controller, uint64_t compaction_needed_bytes,
     const MutableCFOptions& mutable_cf_options,
@@ -922,7 +929,11 @@ std::unique_ptr<WriteControllerToken> ColumnFamilyData::DynamicSetupDelay(
     write_rate = kMinWriteRate;
   }
 
-  return write_controller->GetDelayToken(write_rate);
+  // GetDelayToken returns a DelayWriteToken and also sets the
+  // delayed_write_rate_ which is used by all cfs and dbs using this write
+  // controller. because it also sets the delayed_write_rate_, we need to set
+  // only the smallest active delay request.
+  return write_controller->GetDelayToken(UpdateCFRate(id_, write_rate));
 }
 
 std::pair<WriteStallCondition, ColumnFamilyData::WriteStallCause>
@@ -1048,7 +1059,7 @@ WriteStallCondition ColumnFamilyData::RecalculateWriteStallConditions(
   auto write_stall_condition = WriteStallCondition::kNormal;
   if (current_ != nullptr) {
     auto* vstorage = current_->storage_info();
-    auto write_controller = column_family_set_->write_controller_;
+    auto write_controller = column_family_set_->write_controller_ptr();
     uint64_t compaction_needed_bytes =
         vstorage->estimated_compaction_needed_bytes();
 
@@ -1065,13 +1076,16 @@ WriteStallCondition ColumnFamilyData::RecalculateWriteStallConditions(
 
     // GetWriteStallConditionAndCause returns the first condition met, so its
     // possible that a later condition will require a harder rate limiting.
-    // calculate all conditions with DynamicSetupDelay and reavaluate the
+    // calculate all conditions with DynamicSetupDelay and reevaluate the
     // write_stall_cause. this is only relevant in the kDelayed case.
-    if (dynamic_delay &&
-        write_stall_condition == WriteStallCondition::kDelayed) {
-      write_controller_token_ =
-          DynamicSetupDelay(write_controller, compaction_needed_bytes,
-                            mutable_cf_options, write_stall_cause);
+    if (dynamic_delay) {
+      if (write_stall_condition == WriteStallCondition::kDelayed) {
+        write_controller_token_ =
+            DynamicSetupDelay(write_controller, compaction_needed_bytes,
+                              mutable_cf_options, write_stall_cause);
+      } else {
+        column_family_set_->DeleteSelfFromMapAndMaybeUpdateDelayRate(id_);
+      }
     }
 
     if (write_stall_condition == WriteStallCondition::kStopped &&
@@ -1207,7 +1221,7 @@ WriteStallCondition ColumnFamilyData::RecalculateWriteStallConditions(
       // If the DB recovers from delay conditions, we reward with reducing
       // double the slowdown ratio. This is to balance the long term slowdown
       // increase signal.
-      if (needed_delay) {
+      if (needed_delay && !dynamic_delay) {
         uint64_t write_rate = write_controller->delayed_write_rate();
         write_controller->set_delayed_write_rate(static_cast<uint64_t>(
             static_cast<double>(write_rate) * kDelayRecoverSlowdownRatio));
@@ -1663,16 +1677,14 @@ FSDirectory* ColumnFamilyData::GetDataDir(size_t path_id) const {
   return data_dirs_[path_id].get();
 }
 
-ColumnFamilySet::ColumnFamilySet(const std::string& dbname,
-                                 const ImmutableDBOptions* db_options,
-                                 const FileOptions& file_options,
-                                 Cache* table_cache,
-                                 WriteBufferManager* _write_buffer_manager,
-                                 WriteController* _write_controller,
-                                 BlockCacheTracer* const block_cache_tracer,
-                                 const std::shared_ptr<IOTracer>& io_tracer,
-                                 const std::string& db_id,
-                                 const std::string& db_session_id)
+ColumnFamilySet::ColumnFamilySet(
+    const std::string& dbname, const ImmutableDBOptions* db_options,
+    const FileOptions& file_options, Cache* table_cache,
+    WriteBufferManager* _write_buffer_manager,
+    std::shared_ptr<WriteController> _write_controller,
+    BlockCacheTracer* const block_cache_tracer,
+    const std::shared_ptr<IOTracer>& io_tracer, const std::string& db_id,
+    const std::string& db_session_id)
     : max_column_family_(0),
       file_options_(file_options),
       dummy_cfd_(new ColumnFamilyData(
@@ -1692,6 +1704,7 @@ ColumnFamilySet::ColumnFamilySet(const std::string& dbname,
   // initialize linked list
   dummy_cfd_->prev_ = dummy_cfd_;
   dummy_cfd_->next_ = dummy_cfd_;
+  write_controller_->AddToDbRateMap(&cf_id_to_write_rate_);
 }
 
 ColumnFamilySet::~ColumnFamilySet() {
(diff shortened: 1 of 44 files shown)


db/write_controller.h
```diff
@@ -50,6 +52,16 @@ class WriteController {
...
+  // sets all enties to: is_min = false
+  // set is_min = true for the min rate entry;
+  // returns the min rate from cf_id_to_write_rate_ .
+  // Requires: cf_id_to_write_rate_ is not empty
+  uint64_t GetMinRate();
+
+  // when a previous min rate cf has been removed, we need to set a new min
+  void SetMinRate();
+
+ public:
```
Yuval-Ariel, line 64: this slipped. i'll remove it

db/column_family.cc
```diff
@@ -1694,6 +1773,10 @@ ColumnFamilySet::~ColumnFamilySet() {
   bool dummy_last_ref __attribute__((__unused__));
   dummy_last_ref = dummy_cfd_->UnrefAndTryDelete();
   assert(dummy_last_ref);
+  if (write_controller_) {
```
ayulas, line 1776: its not the right way to protect control map. the lock is private and it should be as part of the RemoveFromDbRateMap function

db/column_family.cc
```diff
@@ -1681,6 +1756,10 @@ ColumnFamilySet::ColumnFamilySet(const std::string& dbname,
   // initialize linked list
   dummy_cfd_->prev_ = dummy_cfd_;
   dummy_cfd_->next_ = dummy_cfd_;
+  if (write_controller_) {
```
ayulas, line 1759: its not the right way to protect control map. the lock is private and it should be as part of the AddToDbRateMap function

db/column_family.cc
```diff
@@ -749,6 +749,9 @@ void ColumnFamilyData::SetDropped() {
   // can't drop default CF
   assert(id_ != 0);
   dropped_ = true;
+  if (column_family_set_->write_controller_->is_dynamic_delay()) {
```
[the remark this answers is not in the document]
Yuval-Ariel: there is an assumption that the write_controller_ in CFset is never null even prior to these changes. e.g. the ptr to write controller is accessed via column_family_set_->write_controller_  in db/column_family.cc ColumnFamilyData::RecalculateWriteStallConditions . 

can we enforce and make sure write_contoller_ is never null or should we go to all these accesses and protect them?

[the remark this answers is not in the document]
Yuval-Ariel: decided to document that there is an assumption of non nullptr. in options.h

db/write_controller.h
```diff
@@ -100,17 +127,27 @@ class WriteController {
...
+  std::atomic<uint64_t> delayed_write_rate_;
+
   // Whether Speedb's dynamic delay is used
   bool dynamic_delay_;
 
+  std::mutex mu_for_map_;
+  std::unordered_set<CfIdToRateMap*> db_id_to_write_rate_map;
+
+  std::condition_variable stop_cv_;
+  // The mutex used by switch_cv_. mutex_ should be acquired beforehand.
```
ayulas, line 148: stop_cv

db/write_controller.cc
```diff
@@ -106,6 +190,8 @@ uint64_t WriteController::NowMicrosMonotonic(SystemClock* clock) {
 StopWriteToken::~StopWriteToken() {
   assert(controller_->total_stopped_ >= 1);
   --controller_->total_stopped_;
+  std::unique_lock<std::mutex> lock(controller_->stop_mutex_);
```
ayulas, line 193: pls create a function in the write controller like you did in the WaitCV so you dont access the controller stop_mutex_ abd cf directly

db/db_impl/db_impl_write.cc
```diff
@@ -1794,7 +1794,15 @@ Status DBImpl::DelayWrite(uint64_t num_bytes,
       // fail any pending writers with no_slowdown
       write_thread_.BeginWriteStall();
       TEST_SYNC_POINT("DBImpl::DelayWrite:Wait");
-      bg_cv_.Wait();
+      // check here if another db is able to wake all the others.
```
ayulas, line 1797: before the change the bg_cv_ wait under the db mutex.. now you release it and wait on the write controller stop cv. whats happened to the previous protection?. the db mutex is not relevant tp the global write controller cv which handles multi db, but i just want to make sure we didnt forget some assumption

db/column_family.cc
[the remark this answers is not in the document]
Yuval-Ariel: why not? its always going to be a valid pointer

db/column_family.cc
```diff
@@ -1136,7 +1073,7 @@ WriteStallCondition ColumnFamilyData::RecalculateWriteStallConditions(
             DynamicSetupDelay(write_controller, compaction_needed_bytes,
                               mutable_cf_options, write_stall_cause);
       } else {
-        MaybeRemoveSelfAndRefreshDelayRate();
+        column_family_set_->DeleteSelfFromMapAndMaybeUpdateDelayRate(id_);
```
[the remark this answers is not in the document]
Yuval-Ariel: thanks but i'll stay with the more informative name

db/column_family.cc
```diff
@@ -749,6 +749,9 @@ void ColumnFamilyData::SetDropped() {
   // can't drop default CF
   assert(id_ != 0);
   dropped_ = true;
+  if (column_family_set_->write_controller_->is_dynamic_delay()) {
+    column_family_set_->DeleteSelfFromMapAndMaybeUpdateDelayRate(id_);
```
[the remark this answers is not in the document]
Yuval-Ariel: mentioned above

db/column_family.cc
```diff
@@ -1760,6 +1774,21 @@ ColumnFamilyData* ColumnFamilySet::CreateColumnFamily(
...
 
+// returns the min rate to set
+uint64_t ColumnFamilySet::UpdateCFRate(uint32_t id, uint64_t write_rate) {
+  return write_controller_->InsertToMapAndGetMinRate(id, &cf_id_to_write_rate_,
+                                                     write_rate);
+}
+
+// Removes the cf from the cf_id_to_write_rate_ map if it exists and if this cf
+// was the one with the min write rate then find and set a new min.
+void ColumnFamilySet::DeleteSelfFromMapAndMaybeUpdateDelayRate(uint32_t id) {
```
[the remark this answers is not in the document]
Yuval-Ariel: mentioned above

db/write_controller.cc
```diff
@@ -33,6 +38,82 @@ std::unique_ptr<WriteControllerToken> WriteController::GetDelayToken(
...
+  uint64_t min_rate = delayed_write_rate();
+  auto cf_rate = (*cf_map)[id];
+  // the cf is already in the map so it shouldnt be possible for it to have a
+  // lower rate than the delayed_write_rate_ unless set_max_delayed_write_rate
+  // has been used which also sets delayed_write_rate_
+  // its fine for several cfs to have the same min_rate.
+  return cf_rate <= min_rate;
+}
+
+void WriteController::DeleteSelfFromMapAndMaybeUpdateDelayRate(
```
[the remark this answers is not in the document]
Yuval-Ariel: why?

db/db_impl/db_impl_write.cc
Yuval-Ariel: the bg_cv_ unlocks the db mutex once it starts waiting in any case. i didnt want to keep holding it while waiting on the write controller mutex

db/write_controller.h
Yuval-Ariel: all members end with _. its the db_id_to_write_rate_map that needs to be changed. done.

db/compaction/compaction_job.cc
```diff
@@ -1983,7 +1983,7 @@ Env::IOPriority CompactionJob::GetRateLimiterPriority() {
   if (versions_ && versions_->GetColumnFamilySet() &&
       versions_->GetColumnFamilySet()->write_controller()) {
     WriteController* write_controller =
-        versions_->GetColumnFamilySet()->write_controller();
+        versions_->GetColumnFamilySet()->write_controller().get();
```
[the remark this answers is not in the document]
Yuval-Ariel: why?

db/write_controller.cc
```diff
@@ -56,9 +137,10 @@ uint64_t WriteController::GetDelay(SystemClock* clock, uint64_t num_bytes) {
     return 0;
   }
 
-  if (credit_in_bytes_ >= num_bytes) {
-    credit_in_bytes_ -= num_bytes;
-    return 0;
+  auto credits = credit_in_bytes_.load();
```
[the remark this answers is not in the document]
Yuval-Ariel: i wanted to keep the original functionality (old lines 59-61) but with multithreading protection and without locking the write_controller mutex. credits will automatically update its value to the value in credit_in_bytes_ if it fails the comparison in each iteration. 
the db_mutex is indeed locked but credit_in_bytes_ should be protected from other **dbs** using the same write_controller.

db/write_controller.cc
Yuval-Ariel: done

db/column_family.cc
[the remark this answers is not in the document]
ayulas: cause it is a member of the set and it should be private. you already changed this in various places after the first comments i gave you. this was left

db/compaction/compaction_job.cc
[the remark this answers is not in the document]
ayulas: you dont access the set write controller directly, so if you want to get pointer and not shared ptr create an access function to it.. its the write way c++ way

db/write_controller.cc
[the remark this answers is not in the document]
ayulas: i think its nicer and make the code cleaner. but its your call.. any way the function name is too long...

db/column_family.cc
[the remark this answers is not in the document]
Yuval-Ariel: oops sry didnt notice it was write_controller_ instead of write_controller()..

ayulas (dismissed):

ayulas (dismissed):

db/write_controller.cc
```diff
@@ -33,6 +38,84 @@ std::unique_ptr<WriteControllerToken> WriteController::GetDelayToken(
...
+  if (cf_write_rate <= delayed_write_rate()) {
+    return cf_write_rate;
+  } else if (was_min) {
+    return GetMapMinRate();
+  } else {
+    return delayed_write_rate();
+  }
+}
+
+void WriteController::WaitOnCV(const ErrorHandler& error_handler) {
```
ayulas, line 109: the right way is to 
    {
      std::unique_lock <<std::mutex>> lock(stop_mutex_);
      while (error_handler.GetBGError().ok() && IsStopped()) {
        stop_cv_.wait(lock);
      }
    }

db/write_controller.cc
```diff
@@ -103,9 +187,15 @@ uint64_t WriteController::NowMicrosMonotonic(SystemClock* clock) {
   return clock->NowNanos() / std::milli::den;
 }
 
+void WriteController::NotifyCV() {
+  std::lock_guard<std::mutex> lock(stop_mutex_);
```
ayulas, line 191: why you changed from std::unique_lock to std::lock_guard.  pls change it back
do the total_stopped_ update in its scope and the notify out the lock scope

db/write_controller.cc
```diff
@@ -103,9 +187,15 @@ uint64_t WriteController::NowMicrosMonotonic(SystemClock* clock) {
...
+  --total_stopped_;
+  stop_cv_.notify_all();
+}
+
 StopWriteToken::~StopWriteToken() {
   assert(controller_->total_stopped_ >= 1);
```
ayulas, line 197: pls move this assert to the NotifyCV function. you access total_stopped_ only in the WriteController member function except this.... the right way is to access it internally. and it should be private

db/write_controller.cc
Yuval-Ariel: std::lock_guard is sufficient in this case. why do you think unique is better?

db/write_controller.cc
Yuval-Ariel: done

db/write_controller.cc
Yuval-Ariel: its not code we wrote so i dont think we should change what doesnt really needs to be changed but sure..

ayulas (approved):

ayulas (approved):

Yuval-Ariel:
@erez-speedb , plz check that theres no deg in normal suite

erez-speedb:
@Yuval-Ariel  where is the issue for this PR?

Yuval-Ariel:
#346 @erez-speedb

PR #397, opened 2023-02-10
Add useBaseAddress to OptionTypeInfo

mrambacher:
(This replaces #308, which will be abandoned due to git branching issues).

When an OptionsTypeInfo is defined, it is passed the offset of the field in question. This offset is the relative location of the field in the structure being analyzed. This PR adds a "useBaseAddress" flag that tells the option system to use the base address rather than the offset to any custom functions.  The flag is only used if a custom function is used and is ignored if using the default behaviors.

This new flag is useful when there are relationships between fields in the structure. For example:

If a std::shared_ptr shared_env value is added to the DBOptions, the "shared" and "env" can be marked as useBase. This would allow the system to serialize either of the values (either shared or env and not both) and allow the Create method to update the other value.
For things like Compaction, the different Options (CompactionOptionsFIFO/Universal) could be skipped during serialization if they did not apply (so if FIFO was specified, do not store or validate Universal options).
Fields that are specific to certain implementations could be moved/associated with the implementation.
When multiple fields are validated/prepared/sanitized together, the code can be placed in the OptionsTypeInfo rather than writing class-specific implementations of these methods. For example, the BlockBasedTableOptions have fields that are related. Using this new flag allows the relationship to be initialized and validated without the class methods.


table/block_based/block_based_table_factory.cc
```diff
@@ -280,36 +280,167 @@ static std::unordered_map<std::string, OptionTypeInfo>
           OptionType::kChecksumType, OptionVerificationType::kNormal,
           OptionTypeFlags::kNone}},
         {"no_block_cache",
-         {offsetof(struct BlockBasedTableOptions, no_block_cache),
-          OptionType::kBoolean, OptionVerificationType::kNormal,
-          OptionTypeFlags::kNone}},
+         OptionTypeInfo(offsetof(struct BlockBasedTableOptions, no_block_cache),
```
RoyBenMoshe, line 283: * I see in this file some differences from the old PR, like lines 259-261 that were deleted in the old PR, is it ok?
* Did you test the new functionality? How did you test it?

mrambacher: Should be fixed.

I tested this code by running the unit tests.

RoyBenMoshe: * I think that might be some more differences in addition to the one above, please check it.
* Can you please elaborate how the new functionality(the definition and the use of the relations between fields) is tested with the current unit tests, aren't we want to add specific tests for it?

udi-speedb:
@mrambacher - Could you please comment / reply / handle the last review comments @RoyBenMoshe gave?

table/block_based/block_based_table_factory.cc
mrambacher: > * I think that might be some more differences in addition to the one above, please check it.

I think that the differences should be resolved now.

> * Can you please elaborate how the new functionality(the definition and the use of the relations between fields) is tested with the current unit tests, aren't we want to add specific tests for it?

I added a unit test to do this relationship explicitly

udi-speedb (dismissed):

table/block_based/block_based_table_factory.cc
```diff
@@ -434,23 +562,11 @@ static std::unordered_map<std::string, OptionTypeInfo>
...
-        std::make_shared<CacheReservationManagerImpl<
-            CacheEntryRole::kBlockBasedTableReader>>(
-            table_options_.block_cache)));
-  }
-}
-
-void BlockBasedTableFactory::InitializeOptions() {
+  // Initialize/Prepare the BlockBasedTableOptions
+  // Note that comparable code is also implemented in the OptionTypeMap;
+  // the code is needed here as well in order to support LITE mode
```
ofriedma, line 569: RocksDB LITE has been removed

ofriedma (changes requested):

table/block_based/block_based_table_factory.cc
```diff
@@ -505,11 +622,26 @@ void BlockBasedTableFactory::InitializeOptions() {
       options_overrides_iter->second.charged = options.charged;
     }
   }
+  //**TODO: Move this code into PrepareOptions
```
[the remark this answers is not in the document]
mrambacher: @ofriedma This file has been reverted to match that of main.  I am not sure why it is still showing up as different

table/block_based/block_based_table_factory.cc
mrambacher: > RocksDB LITE has been removed

This file has not changed from main.  Not sure why it is in this review any longer.

ofriedma (approved):

ofriedma (approved):

udi-speedb:
@mrambacher, @ofriedma - Please handle this PR so we may merge it.
A reminder: It's now mandatory to resolve all of the comments (the reviewer should resolve) to be able to merge.
Thanks

ofriedma:
@Yuval-Ariel ready to be merge

udi-speedb:
@mrambacher - Please fix the license / History issue. Thanks

ofriedma (approved):

udi-speedb:
@or - Please review, resolve applicable comments, and indicate when ready to merge. Thanks

ofriedma (approved):

PR #398, opened 2023-02-13
dont use AVX512 with asan

Yuval-Ariel:


Yuval-Ariel:
the problem doesnt reproduce with this branch

udi-speedb:
@Yuval-Ariel Please add a comment explaining why you avoid using AVX512 when MUST_FREE_HEAP_ALLOCATIONS (== ASAN).

Yuval-Ariel:
> @Yuval-Ariel Please add a comment explaining why you avoid using AVX512 when MUST_FREE_HEAP_ALLOCATIONS (== ASAN).

right you are, done

udi-speedb (approved):

mrambacher:
Do we understand why this happens on the Speedb code and is not seen in Rocks?  Is it a different test suite or something different?

udi-speedb:
> Do we understand why this happens on the Speedb code and is not seen in Rocks? Is it a different test suite or something different?

Unless I am mistaken, it does happen in RocksDB as well. This is a workaround that would let us test with ASAN. We should open an issue for RocksDB.
@Yuval-Ariel - Please confirm or correct me as applicable. Thanks

Yuval-Ariel:
yes, @udi-speedb  is correct.

Issue #394, opened 2023-02-08
TTL - Don't return expired objects


bosmatt:
Currently the Get operation returns values that their ttl has expired. 
The requirement is to remove these objects from the Get output.

ofriedma:
db_bench should work now

ofriedma:
The test ran far more time than configured, I'm investigating it

erez-speedb:
Run db_bench, verified the TTL cause DB reduction.
There is no actual perf test to compare, just check it won hang.
Moving to review.

PR #403, opened 2023-02-15
Skip expired object while using DBWithTtl

ofriedma:
Currently even if objects have been expired we still return them until
compaction delete those objects.

This commit adds a skip_expired_data to the ReadOptions, if this option
is true, it Get, MultiGet and iterator will not return objects that have been
expired by the configured ttl.
Currently this feature does not include support of for Merge
operation.
For Read Only and iterator pinned data the skip_expired_data will be enforced and expired keys
will not be returned.

Resolves: https://github.com/speedb-io/speedb/issues/394 , https://github.com/speedb-io/speedb/issues/417

### DIFF
--- a/build_tools/check-sources.sh
+++ b/build_tools/check-sources.sh
@@ -31,7 +31,8 @@ fi
 
 git grep -n 'using namespace' -- ':!build_tools' ':!docs' \
     ':!third-party/folly/folly/lang/Align.h' \
-    ':!third-party/gtest-1.8.1/fused-src/gtest/gtest.h'
+    ':!third-party/gtest-1.8.1/fused-src/gtest/gtest.h' \
+    ':!examples/speedb_with_ttl_example.cc'
 if [ "$?" != "1" ]; then
   echo '^^^^ Do not use "using namespace"'
   BAD=1
--- a/db_stress_tool/db_stress_test_base.cc
+++ b/db_stress_tool/db_stress_test_base.cc
@@ -682,6 +682,12 @@ void StressTest::OperateDb(ThreadState* thread) {
   read_opts.async_io = FLAGS_async_io;
   read_opts.adaptive_readahead = FLAGS_adaptive_readahead;
   read_opts.readahead_size = FLAGS_readahead_size;
+  if (gflags::GetCommandLineFlagInfoOrDie("ttl").is_default &&
+      FLAGS_skip_expired_data && FLAGS_ttl < 1) {
+    auto error_msg =
+        IOStatus::InvalidArgument("skip_expired_data must be set with ttl");
+  }
+  read_opts.skip_expired_data = FLAGS_skip_expired_data;
   WriteOptions write_opts;
   if (FLAGS_rate_limit_auto_wal_flush) {
     write_opts.rate_limiter_priority = Env::IO_USER;
@@ -2492,7 +2498,6 @@ void StressTest::Open(SharedState* shared) {
 
   Status s;
 
-  if (FLAGS_ttl == -1) {
     std::vector<std::string> existing_column_families;
     s = DB::ListColumnFamilies(DBOptions(options_), FLAGS_db,
                                &existing_column_families);  // ignore errors
@@ -2611,11 +2616,44 @@ void StressTest::Open(SharedState* shared) {
 #endif  // !ROCKSDB_LITE
         {
           if (db_preload_finished_.load() && FLAGS_read_only) {
-            s = DB::OpenForReadOnly(DBOptions(options_), FLAGS_db,
-                                    cf_descriptors, &column_families_, &db_);
+            if (FLAGS_ttl == -1) {
+              s = DB::OpenForReadOnly(DBOptions(options_), FLAGS_db,
+                                      cf_descriptors, &column_families_, &db_);
+            } else {
+              DBWithTTL* dbttl;
+              std::vector<int32_t> ttls;
+              for (size_t i = 0; i < cf_descriptors.size(); ++i) {
+                ttls.push_back(FLAGS_ttl);
+              }
+              s = DBWithTTL::Open(DBOptions(options_), FLAGS_db, cf_descriptors,
+                                  &column_families_, &dbttl, ttls, true);
+              if (!s.ok()) {
+                fprintf(stderr, "Cannot read only open db with ttl. %s\n",
+                        s.ToString().c_str());
+                exit(1);
+              }
+              db_ = dbttl;
+            }
           } else {
-            s = DB::Open(DBOptions(options_), FLAGS_db, cf_descriptors,
-                         &column_families_, &db_);
+            if (FLAGS_ttl == -1) {
+              s = DB::Open(DBOptions(options_), FLAGS_db, cf_descriptors,
+                           &column_families_, &db_);
+            } else {
+              std::vector<int32_t> ttls;
+              for (size_t i = 0; i < cf_descriptors.size(); ++i) {
+                ttls.push_back(FLAGS_ttl);
+              }
+              DBWithTTL* dbttl;
+
+              s = DBWithTTL::Open(DBOptions(options_), FLAGS_db, cf_descriptors,
+                                  &column_families_, &dbttl, ttls);
+              if (!s.ok()) {
+                fprintf(stderr, "Cannot open db with ttl. %s\n",
+                        s.ToString().c_str());
+                exit(1);
+              }
+              db_ = dbttl;
+            }
           }
         }
 
@@ -2743,16 +2781,6 @@ void StressTest::Open(SharedState* shared) {
       exit(1);
 #endif  // !ROCKSDB_LITE
     }
-  } else {
-#ifndef ROCKSDB_LITE
-    DBWithTTL* db_with_ttl;
-    s = DBWithTTL::Open(options_, FLAGS_db, &db_with_ttl, FLAGS_ttl);
-    db_ = db_with_ttl;
-#else
-    fprintf(stderr, "TTL is not supported in LITE mode\n");
-    exit(1);
-#endif
-  }
 
   if (FLAGS_preserve_unverified_changes) {
     // Up until now, no live file should have become obsolete due to these
--- a/examples/CMakeLists.txt
+++ b/examples/CMakeLists.txt
@@ -43,3 +43,13 @@ add_executable(multi_processes_example
   multi_processes_example.cc)
 target_link_libraries(multi_processes_example
   ${ROCKSDB_LIB})
+
+add_executable(speedb_with_ttl_example
+  speedb_with_ttl_example.cc)
+  target_link_libraries(speedb_with_ttl_example
+    ${ROCKSDB_LIB})
+
+add_executable(speedb_is_awesome_example
+  speedb_is_awesome_example.cc)
+  target_link_libraries(speedb_is_awesome_example
+    ${ROCKSDB_LIB})
--- a/examples/Makefile
+++ b/examples/Makefile
@@ -19,7 +19,8 @@ CFLAGS += -Wstrict-prototypes
 
 .PHONY: clean static_lib
 
-all: simple_example column_families_example compaction_filter_example compact_files_example c_simple_example optimistic_transaction_example transaction_example options_file_example rocksdb_backup_restore_example speedb_is_awesome_example
+all: simple_example column_families_example compaction_filter_example compact_files_example c_simple_example optimistic_transaction_example \
+	 transaction_example options_file_example rocksdb_backup_restore_example speedb_is_awesome_example speedb_with_ttl_example
 
 simple_example: static_lib simple_example.cc
 	$(CXX) $(CXXFLAGS) $@.cc -o$@ ../$(LIBNAME).a -I../include -O2 -std=c++17 $(PLATFORM_LDFLAGS) $(PLATFORM_CXXFLAGS) $(EXEC_LDFLAGS)
@@ -54,11 +55,16 @@ multi_processes_example: static_lib multi_processes_example.cc
 speedb_is_awesome_example: static_lib speedb_is_awesome_example.cc
 	$(CXX) $(CXXFLAGS) $@.cc -o$@ ../$(LIBNAME).a -I../include -O2 -std=c++17 $(PLATFORM_LDFLAGS) $(PLATFORM_CXXFLAGS) $(EXEC_LDFLAGS)
 
+speedb_with_ttl_example: static_lib speedb_with_ttl_example.cc
+	$(CXX) $(CXXFLAGS) $@.cc -o$@ ../$(LIBNAME).a -I../include -O2 -std=c++17 $(PLATFORM_LDFLAGS) $(PLATFORM_CXXFLAGS) $(EXEC_LDFLAGS)
+
 rocksdb_backup_restore_example: static_lib rocksdb_backup_restore_example.cc
 	$(CXX) $(CXXFLAGS) $@.cc -o$@ ../$(LIBNAME).a -I../include -O2 -std=c++17 $(PLATFORM_LDFLAGS) $(PLATFORM_CXXFLAGS) $(EXEC_LDFLAGS)
 
 clean:
-	rm -rf ./simple_example ./column_families_example ./compact_files_example ./compaction_filter_example ./c_simple_example c_simple_example.o ./optimistic_transaction_example ./transaction_example ./options_file_example ./multi_processes_example ./rocksdb_backup_restore_example ./speedb_is_awesome_example
+	rm -rf ./simple_example ./column_families_example ./compact_files_example ./compaction_filter_example ./c_simple_example c_simple_example.o \
+		   ./optimistic_transaction_example ./transaction_example ./options_file_example ./multi_processes_example ./rocksdb_backup_restore_example \
+		   ./speedb_is_awesome_example ./speedb_with_ttl_example
 
 static_lib:
 	LIBNAME="$(LIBNAME)" $(MAKE) -C .. static_lib
--- a/examples/speedb_with_ttl_example.cc
+++ b/examples/speedb_with_ttl_example.cc
@@ -1,0 +1,133 @@
+// Copyright (C) 2023 Speedb Ltd. All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//   http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include <unistd.h>
+
+#include <iostream>
+
+#include "rocksdb/db.h"
+#include "rocksdb/options.h"
+#include "rocksdb/slice.h"
+#include "rocksdb/utilities/db_ttl.h"
+
+using namespace ROCKSDB_NAMESPACE;
+
+#if defined(OS_WIN)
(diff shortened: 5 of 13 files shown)


examples/speedb_with_ttl_example.cc
```diff
@@ -0,0 +1,75 @@
...
+#include "rocksdb/slice.h"
+#include "rocksdb/utilities/db_ttl.h"
+
+using ROCKSDB_NAMESPACE::DB;
+using ROCKSDB_NAMESPACE::DBWithTTL;
+using ROCKSDB_NAMESPACE::Options;
+using ROCKSDB_NAMESPACE::ReadOptions;
+using ROCKSDB_NAMESPACE::Status;
+using ROCKSDB_NAMESPACE::WriteOptions;
+
```
udi-speedb, line 30: It may be cleaner, and therefore preferable, to just add a `using namespace ROCKSDB_NAMESPACE` as this is what we want here anyway

udi-speedb:
Does the partial / hybrid support (Get / MultiGet strict ttl while iterator / merge non-strict ttl) make sense for users? 
I understand that compaction is a background process and, therefore, the user has no way of knowing when the keys would be compacted (and consequently removed), but in this PR a user may issue a Get() -> No key, Iter() -> Key returns, Get() -> No key, which will not happen in non-strict ttl.

examples/speedb_with_ttl_example.cc
ofriedma: Done

udi-speedb:
Please add the issue's / pr's number at the end of the commit message

utilities/ttl/db_ttl_impl.h
```diff
@@ -201,6 +201,8 @@ class TtlCompactionFilterFactory : public CompactionFilterFactory {
   const Customizable* Inner() const override {
     return user_comp_filter_factory_.get();
   }
+  int32_t GetTtl() { return ttl_; }
+  SystemClock* GetClock() { return clock_; }
```
udi-speedb, line 205: I think this one is not in use as well

include/rocksdb/options.h
```diff
@@ -334,6 +334,10 @@ struct ColumnFamilyOptions : public AdvancedColumnFamilyOptions {
   // Default: nullptr
   std::shared_ptr<SstPartitionerFactory> sst_partitioner_factory = nullptr;
 
+  // When open DB with TTL it will be the time to live of each object
+  // Default: 0
+  int32_t ttl_skip_expired_data = 0;
```
udi-speedb, line 339: Please emphasize that the default (0) means "never-stale"

utilities/ttl/ttl_test.cc
```diff
@@ -726,6 +726,28 @@ TEST_F(TtlTest, DeleteRangeTest) {
...
+TEST_F(TtlTest, SkipExpiredTtlGetTest) {
+  OpenTtl(1);
+  ASSERT_OK(db_ttl_->Put(WriteOptions(), "a", "val"));
+  env_->Sleep(2);
+  auto ropts = ReadOptions();
+  ropts.skip_expired_data = true;
+  std::string value;
+  ASSERT_TRUE(db_ttl_->Get(ropts, "a", &value).IsNotFound());
+  CloseTtl();
+}
```
udi-speedb, line 738: I would add the following to the test:
1. Show that the value is returned when skip_expired_data == false
2. Show that the value is returned before becoming stale for skip = {false, true} (it might make sense to make this test parametrized)
3. Test behavior for the default ttl value of 0
4. Actually sleeping for 2 seconds make the tests long. I suggest using the sleep mocking utilities they have.
5. Showing that a read options with a larger ttl will avoid skipping

build_tools/check-sources.sh
```diff
@@ -31,7 +31,8 @@ fi
 
 git grep -n 'using namespace' -- ':!build_tools' ':!docs' \
     ':!third-party/folly/folly/lang/Align.h' \
-    ':!third-party/gtest-1.8.1/fused-src/gtest/gtest.h'
+    ':!third-party/gtest-1.8.1/fused-src/gtest/gtest.h' \
+    ':!examples/speedb_with_ttl_example.cc'
```
udi-speedb, line 35: Why is this needed here?
Are we missing this for other files we have added?

examples/CMakeLists.txt
```diff
@@ -43,3 +43,15 @@ add_executable(multi_processes_example
...
+add_executable(speedb_with_ttl_example
+  EXCLUDE_FROM_ALL
+  speedb_with_ttl_example.cc)
+  target_link_libraries(speedb_with_ttl_example)
+    $({ROCKSDB_LIB})
+
+add_executable(speedb_is_awesome_examle
+  EXCLUDE_FROM_ALL
+  speedb_is_awesome_examle.cc)
+  target_link_libraries(speedb_is_awesome_examle)
```
udi-speedb, line 56: the target_link_libraries is indented differently (in both speedb_with_ttl_example and speedb_is_awesome_examle) - I am not using cmake. Does this have a real impact on the outcome?
BTW - speedb_is_awesome_examle has a typo - **example**, not **examle**
I must have missed these in my review of @RoyBenMoshe speedb_is_awesome_example PR. 

Did we verify that it runs correctly under cmake?

examples/Makefile
```diff
@@ -19,7 +19,7 @@ CFLAGS += -Wstrict-prototypes
 
 .PHONY: clean static_lib
 
-all: simple_example column_families_example compaction_filter_example compact_files_example c_simple_example optimistic_transaction_example transaction_example options_file_example rocksdb_backup_restore_example speedb_is_awesome_example
+all: simple_example column_families_example compaction_filter_example compact_files_example c_simple_example optimistic_transaction_example transaction_example options_file_example rocksdb_backup_restore_example speedb_is_awesome_example speedb_with_ttl_example
```
udi-speedb, line 22: Could we shorten the line and use multi-lines?

examples/Makefile
```diff
@@ -54,11 +54,14 @@ multi_processes_example: static_lib multi_processes_example.cc
...
 
+speedb_with_ttl_example: static_lib speedb_with_ttl_example.cc
+	$(CXX) $(CXXFLAGS) $@.cc -o$@ ../$(LIBNAME).a -I../include -O2 -std=c++17 $(PLATFORM_LDFLAGS) $(PLATFORM_CXXFLAGS) $(EXEC_LDFLAGS)
+
 rocksdb_backup_restore_example: static_lib rocksdb_backup_restore_example.cc
 	$(CXX) $(CXXFLAGS) $@.cc -o$@ ../$(LIBNAME).a -I../include -O2 -std=c++17 $(PLATFORM_LDFLAGS) $(PLATFORM_CXXFLAGS) $(EXEC_LDFLAGS)
 
 clean:
-	rm -rf ./simple_example ./column_families_example ./compact_files_example ./compaction_filter_example ./c_simple_example c_simple_example.o ./optimistic_transaction_example ./transaction_example ./options_file_example ./multi_processes_example ./rocksdb_backup_restore_example ./speedb_is_awesome_example
+	rm -rf ./simple_example ./column_families_example ./compact_files_example ./compaction_filter_example ./c_simple_example c_simple_example.o ./optimistic_transaction_example ./transaction_example ./options_file_example ./multi_processes_example ./rocksdb_backup_restore_example ./speedb_is_awesome_example ./speedb_with_ttl_example
```
udi-speedb, line 64: Same here. Too long

utilities/ttl/db_ttl_impl.cc
```diff
@@ -170,6 +172,8 @@ void DBWithTTLImpl::SanitizeOptions(int32_t ttl, ColumnFamilyOptions* options,
     options->merge_operator.reset(
         new TtlMergeOperator(options->merge_operator, clock));
   }
+
+  options->ttl_skip_expired_data = ttl;
```
udi-speedb, line 176: This doesn't make sense. You have added it as an option for the user, but overwrite it using the ttl value specified. That means that it's not truly a user-settable option

utilities/ttl/ttl_test.cc
udi-speedb, line 738: And please add a space line between the tests

utilities/ttl/ttl_test.cc
```diff
@@ -726,6 +726,28 @@ TEST_F(TtlTest, DeleteRangeTest) {
...
+  ASSERT_OK(db_ttl_->Put(WriteOptions(), "a", "val"));
+  env_->Sleep(4);
+  auto ropts = ReadOptions();
+  ropts.skip_expired_data = true;
+  std::vector<std::string> values;
+  ASSERT_TRUE(db_ttl_->MultiGet(ropts, {"a"}, &values)[0].IsNotFound());
+  ASSERT_TRUE(values[0].empty());
+  CloseTtl();
+}
+
```
udi-speedb, line 750: Same comments as above, but also:

1. Please also test the mutli-get versions that accept a single cf and multiple cf-s
I suggest adding another cf, but have the 2 cf-s have different ttl-s and verifying that each cf complies with its ttl

udi-speedb (changes requested):

build_tools/check-sources.sh
ofriedma: it is needed because "using namespace" is not valid in our tests, unless the is being added to this script

udi-speedb:
Please add the PR's number (#403) at the end of the commit message.

utilities/ttl/db_ttl_impl.h
```diff
@@ -113,23 +113,69 @@ class DBWithTTLImpl : public DBWithTTL {
...
+  explicit TtlIterator(Iterator* iter, int32_t ttl, bool skip_expired_data,
+                       SystemClock* clock)
+      : iter_(iter),
+        ttl_(ttl),
+        skip_expired_data_(skip_expired_data),
+        clock_(clock)
+
+  {
+    assert(iter_);
+  }
```
udi-speedb, line 125: A comment applicable to this as well as the other methods below.
As a general rule, header files should include the methods' declarations and not their definitions (implementation). Exceptions exists (e.g., templates and short performance-critical inline methods).

utilities/ttl/db_ttl_impl.h
```diff
@@ -113,23 +113,69 @@ class DBWithTTLImpl : public DBWithTTL {
...
 
-  void Prev() override { iter_->Prev(); }
+  void HandleExpired(bool is_next) {
+    if (!skip_expired_data_) {
+      return;
+    }
+    int64_t current_time;
+    clock_->GetCurrentTime(&current_time);
+    while (Valid()) {
```
udi-speedb, line 167: This is recursive and may potentially iterate recursively over the entire contents of the database if I am not mistaken (=> risk of stack overflow or memory exhaustion, depending on the stack size policy in use).
And, another issue worth considering (in the recursive approach) are the repeated calls to GetCurrentTime() which I think is a rather expensive system call). An iterative solution would allow us to get the time once, and use it for the entire iteration while skipping stale keys.

utilities/ttl/db_ttl_impl.h
```diff
@@ -113,23 +113,69 @@ class DBWithTTLImpl : public DBWithTTL {
...
-  void SeekToFirst() override { iter_->SeekToFirst(); }
+  void SeekToFirst() override {
+    iter_->SeekToFirst();
+    HandleExpired(true);
+  }
+
+  void SeekToLast() override {
+    iter_->SeekToLast();
+    HandleExpired(false);
+  }
```
udi-speedb, line 139: SeekToLast (applicable to SeekToFirst) should succeed when there is any content in the iterated entity.
Would it be possible for the caller to call one of these on an entity it knows is not empty and expect the iterator to be valid after the call?
By a caller I am also referring to internal RocksDB code that may use this iterator (assuming there is such code).
I am not saying I am aware of such a scenario, just raising the issue so we consciously answer it.

udi-speedb (changes requested):

udi-speedb:
@ofriedma - I have submitted all of your comments except for the ttl_test.cc file
I will add them in the next review cycle, after you update the existing files following my comment.

utilities/ttl/db_ttl_impl.h
ofriedma: There is no recursion here, i'm calling Iterator functions and not Ttliterator functions

utilities/ttl/db_ttl_impl.h
ofriedma: The TtlIterator will have the same behaviour as before these changes if the user didn't send skip_expired flag.
The user should always call Valid() before accessing an iterator after any lookup operation such as Seek() or Next().

udi-speedb: I was referring to the case where the user sets skip_expired to true.
But, if indeed the user must always call Valid() then it's a non-issue.
By "user" do you also mean internal RocksDB code that uses the iterator (if there is any such code)?

ofriedma: I don't know, but if there is it should be use the lower level impl anyway and the feature is being active only by read options set to true

udi-speedb (approved):

udi-speedb:
@ofriedma - Please add the issue's number to the commit message. Thanks

tools/db_bench_tool.cc
```diff
@@ -1933,6 +1934,12 @@ DEFINE_bool(track_and_verify_wals_in_manifest,
             ROCKSDB_NAMESPACE::Options().track_and_verify_wals_in_manifest,
             "If true, enable WAL tracking in the MANIFEST");
 
+DEFINE_bool(skip_expired_data, false, "If true, will skip keys expired by TTL");
+
+DEFINE_int32(ttl, -1,
+             "Opens the db with this ttl value if this is not -1. "
+             "Carefully specify a large value such that verifications on "
+             "deleted values don't fail");
```
udi-speedb, line 1942: - If ttl <= 0 the ttl is useless. So, better document this and also validate (below, where applicable)
- How will the user know how to "Carefully specify a large value"? What is a large value?
- You should sanitize (validate) these 2 flags. For example, it doesn't make sense to skip expired data if ttl <= 0.

tools/db_bench_tool.cc
```diff
@@ -5338,6 +5346,19 @@ class Benchmark {
...
+          kDefaultColumnFamilyName, ColumnFamilyOptions(options)));
+      DBWithTTL* dbttl;
+      s = DBWithTTL::Open(options, db_name, column_families, &db->cfh, &dbttl,
+                          {(int32_t)FLAGS_ttl_seconds});
+      if (s.ok()) {
+        db->db = dbttl;
+      }
+      db->cfh.resize(1);
+      db->num_created = 1;
+      db->num_hot = 1;
```
udi-speedb, line 5361: - Why are you not supporting ttl / db with ttl when there are multiple column families?
- Why no ttl db when in read only?
- If it is indeed the case, you should add it to the sanitization part. IMO invalid combinations should cause db_bench to abort and not run as the user has explicitly requested for specific behaviour and db_bench is not testing the desired behaviour.
- And, if this code stays, you should refactor somehow the duplicate code in this case and the one below (the else block)

tools/db_bench_tool.cc
```diff
@@ -5338,6 +5346,19 @@ class Benchmark {
...
+    } else if (FLAGS_ttl > 0) {
+      std::vector<ColumnFamilyDescriptor> column_families;
+      column_families.push_back(ColumnFamilyDescriptor(
+          kDefaultColumnFamilyName, ColumnFamilyOptions(options)));
+      DBWithTTL* dbttl;
+      s = DBWithTTL::Open(options, db_name, column_families, &db->cfh, &dbttl,
+                          {(int32_t)FLAGS_ttl_seconds});
+      if (s.ok()) {
+        db->db = dbttl;
+      }
```
udi-speedb, line 5358: If Open fails (!s.ok()) we exit below, but, for better readability IMO, I would execute the 3 lines below only if s.ok().

udi-speedb:
What about ttl support in db_stress?
Does it exist? If not, is it possible to add it?

udi-speedb (changes requested):

tools/db_bench_tool.cc
ofriedma: Actually I wanted to keep the same documentation as the db_stress (this is a copy of line by line).
I will try to change it to be better understandable here and add I will sanitise those.

udi-speedb:
@ofriedma - When you have made up your mind about the ttl support / changes in db_stress please update (following our discussion yesterday). Thanks

ofriedma:
@udi-speedb about your assumption, it was correct, small ttl is causing failure of the test, which means the test has not been suitable to handle expiration and deletion by db_stress, probably never, I'm getting not found error whether with or without skip_expired_data when use small ttl (--ttl=2)

udi-speedb:
> @udi-speedb about your assumption, it was correct, small ttl is causing failure of the test, which means the test has not been suitable to handle expiration and deletion by db_stress, probably never, I'm getting not found error whether with or without skip_expired_data when use small ttl (--ttl=2).
> 
> @udi-speedb what do you think we should do in that case?

Thanks for checking this.
I see 2 options:
1. Leave db_stress as is. It wasn't designed to actually test the ttl feature and, I agree that it was probably never tested by FB with ttl. That means we are not really testing our updated TTL feature except via unit tests.
2. Invest in adding ttl debug support. Probably requires being able to know which keys have expired (either via compaction or simply by being too old and not returned via the Get / iterator's retrieval) expired keys. Also requires careful sync between the db_stress code that compares the queried values to the expired values (potential races with compaction).

ofriedma:
The pull request is ready to be reviewed, will squash it once the review will over.

Thank you

udi-speedb:
If db_stress will fail if a user is really trying to run it with a reasonable ttl, why not just not touch it at all?

utilities/ttl/ttl_test.cc
```diff
@@ -726,6 +726,442 @@ TEST_F(TtlTest, DeleteRangeTest) {
...
+TEST_F(TtlTest, CompactionTTLConsiderLatestMergeTest) {
+  Options options;
+  options.create_if_missing = true;
+  options.merge_operator = MergeOperators::CreateStringAppendOperator();
+  std::string key_1 = "a";
+  std::string put_value = "1";
+  ASSERT_OK(DBWithTTL::Open(options, dbname_, &db_ttl_, 1));
+  auto ropts = ReadOptions();
```
udi-speedb, line 753: Is skip_expired_data left false on purpose?
Do you expect the test's results to depend on this?
Regardless, worth documenting

udi-speedb (changes requested):

db_stress_tool/db_stress_test_base.cc
```diff
@@ -682,6 +682,12 @@ void StressTest::OperateDb(ThreadState* thread) {
   read_opts.async_io = FLAGS_async_io;
   read_opts.adaptive_readahead = FLAGS_adaptive_readahead;
   read_opts.readahead_size = FLAGS_readahead_size;
+  if (gflags::GetCommandLineFlagInfoOrDie("ttl").is_default &&
```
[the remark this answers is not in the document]
ofriedma: will add a check for the ttl with skip_expired_data to be greater than 0

utilities/ttl/ttl_test.cc
ofriedma: @udi-speedb 
I've added documentation and comments, change the name of the strict ttl added a condition to ttl <= 0.
The last thing I need to test is the windows environment will run make check on top of Windows tomorrow.

udi-speedb: OK. Please check with @maxb-io the status of building on Windows. He is very familiar with this subject.

udi-speedb (approved):

utilities/ttl/ttl_test.cc
ofriedma: I ran ttl_test on windows, ran without issues

ofriedma:
@udi-speedb please approve

udi-speedb (approved):

Issue #420, opened 2023-03-09
Fix compilation errors on Windows Build


udi-speedb:
When compiling for in Windows (Visual Studio 17 2022, possibly others as well), there is a minor compilation error introduced by Add index size report per column family #338 
This issue will fix these errors

PR #421, opened 2023-03-09
Replace uint with unsinged int (Windows Build Failure) (#420)

udi-speedb:


ofriedma (approved):

udi-speedb:
@Yuval-Ariel - Please merge this so the user has the fix.

Issue #407, opened 2023-02-22
Speedb_CLI - a query and maintance tool for off-line Speedb databases


Guyme:
Speedb_CLI 
External interactive tool that can open existing databases and query and manipulate them with the following capabilities at first stage.
operational : 
read the database (get key)
write to the database (put key, value)
delete a key
delete a range of keys
range get
support multiget
trigger compaction
create a checkpoint
Compress db or Decompress db
support with read only mode
support ttl
informational:
number of key-value pairs in the database
total key size in the db and per CF
average key size (largest, smallest, average)
list keys and size (maybe content as well) and values
more informational and functional support will be added per request or need.

PR #427, opened 2023-03-15
Adding beezcli

ofriedma:
This Pull request is adding a wrapper around ldb and allow using ldb in interactive manner (it is yet possible to use it non interactive just like ldb).
It adds some new functionalities to ldb such as multiget,
summary of keys and values from dump command, allow to not present values on dump and idump.
Fix the ttl parameter, to allow ttl to get value and run compaction on top of db with ttl.

Resolves: https://github.com/speedb-io/speedb/issues/407

### DIFF
--- a/tools/ldb_cmd.cc
+++ b/tools/ldb_cmd.cc
@@ -107,6 +107,7 @@ const std::string LDBCommand::ARG_PREPOPULATE_BLOB_CACHE =
 const std::string LDBCommand::ARG_DECODE_BLOB_INDEX = "decode_blob_index";
 const std::string LDBCommand::ARG_DUMP_UNCOMPRESSED_BLOBS =
     "dump_uncompressed_blobs";
+const std::string LDBCommand::ARG_INTERACTIVE = "interactive";
 
 const char* LDBCommand::DELIM = " ==> ";
 
@@ -212,6 +213,9 @@ LDBCommand* LDBCommand::SelectCommand(const ParsedParams& parsed_params) {
                           parsed_params.flags);
   } else if (parsed_params.cmd == BatchPutCommand::Name()) {
     return new BatchPutCommand(parsed_params.cmd_params,
+                               parsed_params.option_map, parsed_params.flags);
+  } else if (parsed_params.cmd == MultiGetCommand::Name()) {
+    return new MultiGetCommand(parsed_params.cmd_params,
                                parsed_params.option_map, parsed_params.flags);
   } else if (parsed_params.cmd == ScanCommand::Name()) {
     return new ScanCommand(parsed_params.cmd_params, parsed_params.option_map,
@@ -384,7 +388,8 @@ LDBCommand::LDBCommand(const std::map<std::string, std::string>& options,
       create_if_missing_(false),
       option_map_(options),
       flags_(flags),
-      valid_cmd_line_options_(valid_cmd_line_options) {
+      valid_cmd_line_options_(valid_cmd_line_options),
+      ttl_(-1) {
   auto itr = options.find(ARG_DB);
   if (itr != options.end()) {
     db_path_ = itr->second;
@@ -415,7 +420,9 @@ LDBCommand::LDBCommand(const std::map<std::string, std::string>& options,
 
   is_key_hex_ = IsKeyHex(options, flags);
   is_value_hex_ = IsValueHex(options, flags);
-  is_db_ttl_ = IsFlagPresent(flags, ARG_TTL);
+  ParseIntOption(option_map_, ARG_TTL, ttl_, exec_state_);
+  is_db_ttl_ = ((ttl_ != -1) || IsFlagPresent(flags, ARG_TTL));
+  is_no_value_ = IsFlagPresent(flags, ARG_NO_VALUE);
   timestamp_ = IsFlagPresent(flags, ARG_TIMESTAMP);
   try_load_options_ = IsTryLoadOptions(options, flags);
   force_consistency_checks_ =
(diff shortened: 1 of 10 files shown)


ofriedma:
Please do not review yet as some tests have failed

utilities/ttl/db_ttl_impl.cc
```diff
@@ -592,7 +621,32 @@ Status DBWithTTLImpl::Write(const WriteOptions& opts, WriteBatch* updates) {
...
+  return new TtlIterator(db_->NewIterator(opts, column_family), ttl,
+                         skip_expired, cfopts.env->GetSystemClock().get());
+}
+
+void TtlIterator::HandleExpired(bool move_forward) {
+  if (!skip_expired_data_) {
+    return;
+  }
+  int64_t current_time;
+  clock_->GetCurrentTime(&current_time);
```
ayulas, line 638: the raising q is what the intention of the ttl when we took iterator.
if the expired time is related to the time the iterator was taken, you should take the GetCurrentTime in the iterator constructor and calc based on that value on the iter operation. 
if not , as i assume, and the expired is related to the time you access the iter element in the iterator, the GetCurrentTime should be placed inside the loop and get it on each iter element access , since theoretically, you can skip and get to a valid one but its ttl was expired  until you reach it in the loop....

utilities/ttl/db_ttl_impl.h
```diff
@@ -86,6 +86,8 @@ class DBWithTTLImpl : public DBWithTTL {
 
   static bool IsStale(const Slice& value, int32_t ttl, SystemClock* clock);
 
+  bool IsStale(const Slice& value, ColumnFamilyHandle* column_family);
```
ayulas, line 89: add comment in here

utilities/ttl/ttl_test.cc
```diff
@@ -726,6 +726,217 @@ TEST_F(TtlTest, DeleteRangeTest) {
...
+
+TEST_F(TtlTest, SkipExpiredTtlGetMultiTest) {
+  OpenTtl(1);
+  ASSERT_OK(db_ttl_->Put(WriteOptions(), "a", "val"));
+  env_->Sleep(4);
+  auto ropts = ReadOptions();
+  ropts.skip_expired_data = true;
+  std::vector<std::string> values;
+  ASSERT_TRUE(db_ttl_->MultiGet(ropts, {"a"}, &values)[0].IsNotFound());
+  ASSERT_TRUE(values[0].empty());
```
ayulas, line 856: why you need it... if the MultiGet return not found ofcourse no values...

utilities/ttl/ttl_test.cc
```diff
@@ -726,6 +726,217 @@ TEST_F(TtlTest, DeleteRangeTest) {
...
+  CloseTtl();
+}
+
+TEST_F(TtlTest, DoNotSkipUnExpiredTtlGetTest) {
```
ayulas, line 860: do not skip??? you set skip.. . you get the value cause it wasnt expired

utilities/ttl/ttl_test.cc
```diff
@@ -726,6 +726,217 @@ TEST_F(TtlTest, DeleteRangeTest) {
...
+  ASSERT_OK(db_ttl_->Put(WriteOptions(), "a", "val"));
+  env_->Sleep(1);
+  auto ropts = ReadOptions();
+  ropts.skip_expired_data = true;
+  std::string value;
+  ASSERT_OK(db_ttl_->Get(ropts, "a", &value));
+  CloseTtl();
+}
+
+TEST_F(TtlTest, DoNotSkipReadOnlyTtlMultiGetTest) {
```
ayulas, line 871: again the name do not skip is not what you set... ropts.skip_expired_data = true;

utilities/ttl/ttl_test.cc
```diff
@@ -726,6 +726,217 @@ TEST_F(TtlTest, DeleteRangeTest) {
...
+  ASSERT_OK(DBWithTTL::Open(options, dbname_, &db_ttl_, ttl, true));
+  env_->Sleep(ttl + 1);
+  auto statuses = db_ttl_->MultiGet(ropts, {"a", "b"}, &values);
+  for (auto& status : statuses) {
+    ASSERT_TRUE(status.IsNotFound());
+  }
+  CloseTtl();
+}
+
+TEST_F(TtlTest, DoNotSkipReadOnlyTtlGetTest) {
```
ayulas, line 893: as above

utilities/ttl/ttl_test.cc
```diff
@@ -726,6 +726,217 @@ TEST_F(TtlTest, DeleteRangeTest) {
...
+  ASSERT_OK(db_ttl_->Put(WriteOptions(), handles[1], "a", "val"));
+  env_->Sleep(ttl + 1);
+  ASSERT_TRUE(db_ttl_->Get(ropts, handles[0], "a", &value).IsNotFound());
+  ASSERT_OK(db_ttl_->Get(ropts, handles[1], "a", &value));
+  for (auto& h : handles) {
+    delete h;
+    h = nullptr;
+  }
+}
+
```
ayulas, line 939: you need to add test that do flush/compaction after ttl expired so you see that eventhough you set skip_expired_data=false after flush if you try get the keys were skipped in the flush and werent written to the sst so they wont exist

tools/ldb_cmd.cc
```diff
@@ -500,7 +516,7 @@ void LDBCommand::OpenDB() {
     }
   } else {
     // We successfully opened DB in single column family mode.
-    assert(column_families_.empty());
+    // assert(column_families_.empty());
```
ofriedma, line 519: what to do with this assert

ayulas (dismissed):

ofriedma:
@ayulas i added unit test and more comments other than that the code is the same has reviewed

ayulas (dismissed):

ayulas (approved):

ofriedma:
@ayulas added `#include <stdio.h>` because it had an issue when compiled with clang

ayulas (approved):

ofriedma:
added //clang-format off and //clang-format on around #include <stdio.h> otherwise clang-format will try to put stdio after readline which will lead to compilation error in clang

Guyme (approved):

PR #428, opened 2023-03-15
Introduce a TableMemoryOptions struct

mrambacher:
Add a "TableMemoryOptions" struct to contain the fields that will be used to determine how to pin or cache blocks from an SST file.

This PR is the first step in the CleanMemoryManager/PinnedMemoryManager.  This PR restructures some of the arguments to the Table classes to pass in a TableMemoryOptions (instead of individual fields).  In the next phase, a PinnedMemoryManager will be added to this struct to allow for accounting of the memory in use per column family.

This PR introduces no algorithmic changes.  It is a restructuring to add a new struct/argument to the methods.


db/compaction/compaction_job.cc
```diff
@@ -1074,7 +1077,7 @@ void CompactionJob::ProcessKeyValueCompaction(SubcompactionState* sub_compact) {
   }
 
   // Although the v2 aggregator is what the level iterator(s) know about,
-  // the AddTombstones calls will be propagated down to the v1 aggregator.
+  // the AddTombstones[ calls will be propagated down to the v1 aggregator.
```
udi-speedb, line 1080: Seems like the "[" was typed by mistake

db/forward_iterator.cc
```diff
@@ -79,14 +79,15 @@ class ForwardLevelIterator : public InternalIterator {
 
     ReadRangeDelAggregator range_del_agg(&cfd_->internal_comparator(),
                                          kMaxSequenceNumber /* upper_bound */);
+    TableMemoryOptions tmoptions(/*level=*/-1,
+                                 /*max_file_size_for_l0_meta_pin=*/0, false);
```
udi-speedb, line 83: Please add a comment for the false (bottommost level argument)

db/forward_iterator.cc
```diff
@@ -691,14 +692,16 @@ void ForwardIterator::RebuildIterators(bool refresh_sv) {
       l0_iters_.push_back(nullptr);
       continue;
     }
+    TableMemoryOptions tmoptions(
+        /*level=*/-1, MaxFileSizeForL0MetaPin(sv_->mutable_cf_options), false);
```
udi-speedb, line 696: Same here, please a comment for the false

db/forward_iterator.cc
```diff
@@ -771,15 +774,17 @@ void ForwardIterator::RenewIterators() {
       }
       continue;
     }
+    TableMemoryOptions tmoptions(
+        /*level=*/-1, MaxFileSizeForL0MetaPin(svnew->mutable_cf_options),
+        false);
```
udi-speedb, line 779: And here

db/forward_iterator.cc
```diff
@@ -838,14 +843,16 @@ void ForwardIterator::ResetIncompleteIterators() {
       continue;
     }
     DeleteIterator(l0_iters_[i]);
+    TableMemoryOptions tmoptions(
+        /*level=*/-1, MaxFileSizeForL0MetaPin(sv_->mutable_cf_options), false);
+
```
udi-speedb, line 848: And...

table/table_builder.h
```diff
@@ -30,37 +30,52 @@ namespace ROCKSDB_NAMESPACE {
...
+        max_file_size_for_l0_meta_pin(_max_file_size_for_l0_meta_pin),
+        is_bottommost(_is_bottommost) {}
+  // What level this table/file is on, -1 for "not set, don't know." Used
+  // for level-specific statistics.
+  int level = -1;
+  // Largest L0 file size whose meta-blocks may be pinned (can be zero when
+  // unknown).
+  const size_t max_file_size_for_l0_meta_pin = 0u;
+
+  bool is_bottommost = false;
```
udi-speedb, line 46: Wouldn't 3 values (enum) would be preferable - have a "Don't know" value?
I assume that level==-1 should also imply that the is_bottommost is irrelevant.
If it is true that whenever you know the level you also know if it's the bottommost, and apply the "don't know level => don't know is_bottommost", then this is unnecessary (but should be documented)

db/repair.cc
```diff
@@ -603,9 +604,11 @@ class Repairer {
       // from compaction_job.cc around call to UpdateBoundariesForRange (to
       // handle range tombstones extendingg beyond range of other entries).
       ReadOptions ropts;
+      TableMemoryOptions tmopts(/*level=*/-1,
+                                /*max_file_size_for_l0_meta_pin=*/0, false);
```
udi-speedb, line 608: Please add a commnent for the false here as well as in all other cases below.
Why did you choose in this case to explicitly init rather than just default construct?

udi-speedb (changes requested):

udi-speedb:
@mrambacher - This is a first batch of comments. Please address them and also squash all of the commits so I have a single up to date commit to review. Thanks

udi-speedb (changes requested):

udi-speedb:
@mrambacher - I understand this PR is abandoned. Please close it.

udi-speedb:
@mrambacher - Please decide what to do about this PR. Thanks

mrambacher:
Abandoning this PR as taking a different approach.

PR #437, opened 2023-03-19
added check_license.yml to the automation workflows

maxb-io:


Yuval-Ariel:
will be addressed in https://github.com/speedb-io/speedb/pull/526

PR #438, opened 2023-03-19
Global delay bug fix: check again credits under mutex

Yuval-Ariel:
i encountered a bug while running tests for the dirty memory connection to the global write controller and the bug is as follows:
during WriteController::GetDelay , we check if there are enough credits and return if there are, here:
```
auto credits = credit_in_bytes_.load();
  while (credits >= num_bytes) {
    if (credit_in_bytes_.compare_exchange_weak(credits, credits - num_bytes))
      return 0;
 }
```
if there arent enough credits, we proceed under the mutex but dont check again the condition above which might have changed from a different thread. 
this PR adds this additional check and while at it, rename and reorder some members.

### DIFF
--- a/db/write_controller.cc
+++ b/db/write_controller.cc
@@ -24,7 +24,7 @@ std::unique_ptr<WriteControllerToken> WriteController::GetStopToken() {
 std::unique_ptr<WriteControllerToken> WriteController::GetDelayToken(
     uint64_t write_rate) {
   {
-    std::lock_guard<std::mutex> lock(mu_);
+    std::lock_guard<std::mutex> lock(metrics_mu_);
     if (0 == total_delayed_++) {
       // Starting delay, so reset counters.
       next_refill_time_ = 0;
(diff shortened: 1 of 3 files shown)


db/write_controller.cc
```diff
@@ -148,7 +148,13 @@ uint64_t WriteController::GetDelay(SystemClock* clock, uint64_t num_bytes) {
   // Refill every 1 ms
   const uint64_t kMicrosPerRefill = 1000;
 
-  std::lock_guard<std::mutex> lock(mu_);
+  std::lock_guard<std::mutex> lock(metrics_mu_);
```
ayulas, line 151: the use of credit_in_bytes_ is not right as i can see it... as you wrote above 
"// This function is no longer under db mutex since credit_in_bytes_ is atomic"
 in the first check you does threat it as atomic and do compare_exchange_weak meaning you do change its value not under mutex

but after you take the lock , you changed its value not using atomic operation which is problematic 
the lock actually doesnt protect from you from access credit_in_bytes_...
you need to access credit_in_bytes_ with atomic functions

Yuval-Ariel: right, we'll simply take the mutex before to include the first place where credit_in_bytes_ can change so now only 1 thread can change credit_in_bytes_ simultaneously

ayulas (dismissed):

ayulas (approved):

PR #451, opened 2023-03-26
dynamic delay: fix pending bytes rate calculation

### LINKED ISSUE #431: dynamic delay: fix calculation when range is small
fix CalculateWriteDelayDividerAndMaybeUpdateWriteStallCause when calculating pending compaction bytes. 
use number of steps instead of fixed size steps of 1G.

Yuval-Ariel:
closes #431 
use fixed num delay steps instead of fixing the step size and deducing the num steps.
this makes the code simpler and avoids a bug when the excess bytes are greater than num_steps * step_size which then fails the assert: assert(step_num < num_steps);  - since step_num == num_steps.

### DIFF
--- a/db/column_family.cc
+++ b/db/column_family.cc
@@ -896,9 +896,8 @@ int GetL0ThresholdSpeedupCompaction(int level0_file_num_compaction_trigger,
 }  // namespace
 
 namespace {
-const uint64_t Gb = 1ull << 30;
-const uint64_t Mb = 1ull << 20;
 const int kMemtablePenalty = 10;
+const int kNumPendingSteps = 100;
 }  // namespace
 
 double ColumnFamilyData::TEST_CalculateWriteDelayDivider(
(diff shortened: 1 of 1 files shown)


db/column_family.cc
```diff
@@ -1001,21 +1002,19 @@ ColumnFamilyData::CalculateWriteDelayDividerAndMaybeUpdateWriteStallCause(
...
-    uint64_t num_steps = soft_hard_range / step_size;
-    auto extra_bytes = compaction_needed_bytes - soft_limit;
-    auto step_num = static_cast<uint64_t>(extra_bytes / step_size);
-    assert(step_num < num_steps);
-    if (num_steps > 0) {
-      pending_divider = 1 / (1 - (static_cast<double>(step_num) / num_steps));
-    }
+    uint64_t step_size = ceil(soft_hard_range / kNumPendingSteps);
+    uint64_t extra_bytes = compaction_needed_bytes - soft_limit;
+    assert(extra_bytes > 0);
```
RoyBenMoshe, line 1013: I think this condition is always true

RoyBenMoshe (approved):

db/column_family.cc
Yuval-Ariel: you're right, useless assert indeed. i'll remove it

Yuval-Ariel: removed. plz reapprove @RoyBenMoshe

RoyBenMoshe (dismissed):

Yuval-Ariel:
@RoyBenMoshe  one last time (hopefully)

RoyBenMoshe (approved):

Yuval-Ariel:
passed performance tests - large obj and small db with single disk both. no degradation
https://admin.speedb.io/performance?items=RBXHrqju2327X8XshdBu&items=C9qIa3go368igM6YoR3L&colors=%23F06292&colors=%232e7d32 
https://admin.speedb.io/performance?items=WnH1m2z4tUwLVJ9a4KJY&items=2SDJCDUFoZO60PyAfoCM&colors=%23F06292&colors=%232e7d32

PR #499, opened 2023-05-10
WC: fix for stop while shutting down

Yuval-Ariel:
as part of https://github.com/facebook/rocksdb/pull/10765 , a check of whether the waiting db was shutting_down_ was intoduced. 
The write thread then stops waiting when shutting_down_ is turned on which is the case when CancelAllBackgroundWork is called. But the rebased code now waits forever since #346 replaced the waiting on bg_cv_ to a wait on a write controller condition variable inside WriteController::WaitOnCV. 
The condition variable inside WaitOnCV (stop_cv_) is only notified through the detor of a StopWriteToken.

The write_controller_token_ in each CF is being reset in any case after the call to CancelAllBackgroundWork from  DBImpl::CloseHelper() (detor of DBImpl) so if the test would issue a db Close and then wait for the thread there would be no problem.

also consider fixing by instead of replacing CancelAllBackgroundWork with Close, to add a function right after the call to CancelAllBackgroundWork which notifies the stop_cv_ . this will make it clear to further unit test writers that calling CancelAllBackgroundWork is not sufficient in order to remove the stop condition

### DIFF
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -17,6 +17,9 @@ Based on RocksDB 8.1.1
 * stress test: fix decoding error (#498)
 * db_bench and stress: fix WBM initiation (#510)
 * Sanitize max_num_parallel_flushes in WBM if 0 (#460)
+* WriteController: fix for stop while shutting down (#499)
+Also switch to waiting a sec on the CV each time. This is required since a bg error doesn't signal the CV in the WriteController.
+
 ### Miscellaneous
 * disable failing unit tests and paired bloom filter stress testing
 * version: update Speedb patch version to 2.4.1 (#503)
--- a/db/column_family.h
+++ b/db/column_family.h
@@ -487,6 +487,8 @@ class ColumnFamilyData {
       const MutableCFOptions& mutable_cf_options,
       WriteStallCause& write_stall_cause);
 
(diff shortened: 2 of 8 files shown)


Yuval-Ariel:
@udi-speedb , plz wait with the review since changing CancelAllBackgroundWork to Close seemed to reveal a new issue which is handled in https://github.com/speedb-io/speedb/issues/500 .
setting this PR to on hold until the above issue is fixed

Yuval-Ariel:
@udi-speedb , plz review. issue #500  is still unresolved as it requires further discussion so we can proceed with this PR IMO.

udi-speedb:
@Yuval-Ariel - Approving. 
Just a reminder - Did you go over all the places where CancelAllBackgroundWork() is called to verify that no need to call ResetWriteControllerTokens()?

udi-speedb (approved):

Yuval-Ariel:
> Just a reminder - Did you go over all the places where CancelAllBackgroundWork() is called to verify that no need to call ResetWriteControllerTokens()?

yes i did

Yuval-Ariel:
@erez-speedb , plz check that theres no degradation. thanks

Yuval-Ariel:
performance - no degradation. 
still need to update history file

PR #518, opened 2023-05-21
Display cf names in rolled logs with their options

udi-speedb:


Yuval-Ariel (approved):

udi-speedb:
@Yuval-Ariel - Any reason not to merge this?

PR #526, opened 2023-05-23
CI: new workflow - license and HISTORY check

Yuval-Ariel:


maxb-io (changes requested):
a)
workflow trigger:
Should the workflow be executed on every pull request? 
need to add 
"pull_request:"
after line 7

b)
See available pull_request types here:
https://frontside.com/blog/2020-05-26-github-actions-pull_request/

c)
line 54 should be:
if [[ "$file" == 'HISTORY.md' ]]; then history_not_in=0

The HISTORY.md file is missing the License statement 

d)
- what year it should be in the license?

e)
- is the license statement correct?

f)
- should ci files be excluded?

maxb-io (approved):

Issue #536, opened 2023-06-06
optimize_filters_for_memory option is NOT printed to the log


udi-speedb:
The BlockBasedTableFactory::GetPrintableOptions() prints the block based table options to the log. 
The optimize_filters_for_memory is missing and should be added

PR #537, opened 2023-06-06
Print optimize_filters_for_memory option to the log (#536)

udi-speedb:


udi-speedb:
@ofriedma - I am not sure there is a better way to sanitize the offm (optimize_filters_for_memory) option.
I have asked @mrambacher for advice in the slack channel, however, I think that we may leave it as is (+ add to the history file + gitbook documentation).
Anyway, I'll wait for @mrambacher's advice and see.

ofriedma (approved):
LGTM

ofriedma (approved):
LGTM

PR #544, opened 2023-06-08
Re-enable the DropWrites Test

mrambacher:
I could not reproduce the problem with the DropWrites test.  However, based on the comment and the code, I believe the issue might have been caused by log files counting against the files.  I changed the test to only check/count the SST files -- which should be the only ones affected by DropWrites anyway -- and still do not see the problem.

Since I could not reproduce the problem before, it is hard to say for certain that I have addressed the problem.  Hopefully running in the CI will now pass the tests more consistently.


Yuval-Ariel (approved):
sure lets wait and see

Yuval-Ariel:
@mrambacher  the issue seems to have been fixed. plz add a dummy commit to run the "Check TARGETS file and code format" test

Issue #35, opened 2022-06-28
Snapshot optimization 


bosmatt:
Why: 
Improve performance during read heavy snapshot workload 

What :
	In Read only workload, when taking a snapshot while Read, don’t create a new snap if it hasn’t changed, use the last one. 
        In Read only workload, don’t use mutex when the last snapshot is still valid.

isaac-io:
An experimental implementation is in https://github.com/speedb-io/speedb/tree/exp/snapshot-reuse
Note that this implementation is incomplete and suffers from at least the following two bugs (rephrased from Jira):

> ## `last_snapshot_` is incorrectly cached
>
> A caching of the last taken snapshot was implemented in order to facilitate faster snapshot taking in read-mostly workloads, such as pure select MySQL benchmarks, where MySQL is taking a snapshot for each query, and the global DB mutex serializes the reads so it effectively limits the amount of parallelism possible.
>
> That work allows avoiding taking the mutex in read-only workloads and taking the mutex for much shorter periods in workloads that include writes by reusing the last snapshot if it’s still valid. However, in order to ensure that we always hold a valid reference, the DB instance also takes a reference on that snapshot and keeps it alive until the next snapshot is taken, even if multiple writes happened in between, rendering that snapshot useless (because a snapshot is based on the DB’s sequence number, and every write increments it). Moreover, having a live obsolete snapshot (after everyone who needed it released it) prevents the compaction from discarding old keys, increasing space and write amplification.
>
> We need to release the cached snapshot on writes (at least some part of the write is done under the mutex, so it’s valid to release the snapshot during that phase). However, there is a need to be careful about when to release the snapshot, since in the case of transactions `LastPublishedSequenceNumber()` is used rather than `LastSequenceNumber()`, so we only want to release the cached snapshot after the former is incremented. The fact that multiple write flows are currently implemented in RocksDB (single queue, two queues, unordered writes, etc.), with our own addition thrown into the mix with #23, this is a task that potentially involves touching many parts of the code.


> ## snapshots: `min_uncommitted_` may be overwritten by other transactions when providing `last_snapshot_`
>
> When the last snapshot matches the current sequence number, with the snapshot optimisation we return the same `SnapshotImpl` instance that we provide to other callers. However, `WritePrepared` and `WriteUnprepared` transactions also make use of the `min_uncommitted_` field in `SnapshotImpl` to store their own information about the state of the transaction. This may conflict between different transactions, leading to unexpected results.
>
[...]

PR #547, opened 2023-06-08
Snapshot Optimization (https://github.com/speedb-io/speedb/issues/35)

ofriedma:
Motivation:
The most important information inside a snapshot is its Sequence number, which allows the compaction to know if the key-value should be deleted or not.
The sequence number is being changed when modification happens in the db.
This feature allows the db to take a snapshot without acquiring db mutex when the last snapshot has the same sequence number as a new one.
In transactional db with mostly read operations, it should improve performance when used with multithreaded environment and as well other scenarios of taking large amount of snapshots with mostly read operations.

This Feature must have folly library installed.

In order to cache the snapshots, there is last_snapshot_
(folly::atomic_shared_ptr, lock free atomic_shared_ptr) in order to
access the last_snapshot_ created and point to it.
For every GetSnapshotImpl call (where snapshots are being created), the
function checks if the sequence number is different than last_snapshot_,
if no, it creates new snapshot and inside this snapshot it adds a
reference to last_snapshot_ (the reference is cached_snapshot), so this sequence number will remain inside
SnapshotList (SnapshotList is the list of the snapshots in the system and used in compaction to show which snapshots are being used), if there are still snapshots holding this sequence number. If the sequence number as changed or the last_snapshot_ is nullptr it will create the snapshot while acquiring db_mutex.

For ReleaseSnapshotImpl (deleting a snapshot).
We will unref the last_snapshot_ (using comapre_exchange_weak) and if the refcount becomes 0, it will
call Deleter and remove this snapshot entirely from the SnapshotList and
continue with taking the db mutex.
If there are still references, it will return without taking it out from
the SnapshotList nor taking the db mutex


ofriedma:
@ayulas Ready for review , thank you

db/db_impl/db_impl.cc
```diff
@@ -3711,7 +3711,57 @@ Status DBImpl::GetTimestampedSnapshots(
...
+    if (lock) {
+      mutex_.Unlock();
+    }
+    delete s;
+    return nullptr;
+  }
+  auto snapshot_seq = GetLastPublishedSequence();
+  SnapshotImpl* snapshot =
+      snapshots_.New(s, snapshot_seq, unix_time, is_write_conflict_boundary);
+  SnapshotImpl* user_snapshot = new SnapshotImpl;
```
ayulas, line 3750: why? you have the s... why creating a new object??
you need to add in the snap structure the notion if its an holder or the original snap thats it.
you should use s and not create a new one
also why not adding to the snapshots_.New function the most of the knowledge...
we want to reduce the db mutex as much as we can
first use the list mutex
then the list should be a set order by seq number (since you can have many threads that doing snap while writing is in progress and because you are updating the snap list by lock in a short scope you might insert a snap with older seq after a new one ) then you update the last snap list object by taking the last object in the set (still under the list mutex) doing it by store atomic function and increase the snap count to 1
when you create a new snap after checking if you supported the snap creation at all, load the last snap (atomically) and on it do your checking.

db/db_impl/db_impl.cc
```diff
@@ -3861,7 +3912,74 @@ bool CfdListContains(const CfdList& list, ColumnFamilyData* cfd) {
...
 }  //  namespace
+#ifdef ROCKSDB_SNAP_OPTIMIZATION
+void DBImpl::ReleaseSnapshot(const Snapshot* s) {
+  if (s == nullptr) {
+    // DBImpl::GetSnapshot() can return nullptr when snapshot
+    // not supported by specifying the condition:
+    // inplace_update_support enabled.
+    return;
+  }
+  snapshots_.count_.fetch_sub(1);
```
ayulas, line 3923: no... if its a reference snap it wasnt part of the list

db/db_impl/db_impl.cc
```diff
@@ -3861,7 +3912,74 @@ bool CfdListContains(const CfdList& list, ColumnFamilyData* cfd) {
...
+    delete snapshot;
+  }
+  if (!snapshots_.deleteitem) {
+    return;
+  }
+  {
+    InstrumentedMutexLock l(&mutex_);
+    std::scoped_lock<std::mutex> snaplock(snapshots_.lock);
+    snapshots_.deleteitem = false;
+    uint64_t oldest_snapshot;
```
ayulas, line 3941: exactly the same code as in the original scope. you copied the same code...
i dont think we should divide it at all.

db/snapshot_impl.h
```diff
@@ -42,17 +60,25 @@ class SnapshotImpl : public Snapshot {
...
   uint64_t timestamp_;
 
   // Will this snapshot be used by a Transaction to do write-conflict checking?
   bool is_write_conflict_boundary_;
+#endif
 };
 
 class SnapshotList {
  public:
+#ifdef ROCKSDB_SNAP_OPTIMIZATION
```
ayulas, line 75: i dont think you should divide code. snap list should have mutex, last snap , the delete item is not needed and you can do the code much much simpler / now you have in some parts holes

ofriedma:
Summarize offered changes:

Fix Release and Get so we will always delete when needed.
Separate count and logical count.
Unix time and sequence number should be together.
move ifdef within the releasesnapshot
Try for GetSnapshot as well.
Add new RefSnapshot to db_impl
move New node snapshot_list logic to New.
Add to New another parameter of re-add.

ofriedma:
@ayulas Please review

ofriedma:
ready for review

db/db_impl/db_impl.cc
```diff
@@ -3815,15 +3808,14 @@ DBImpl::CreateTimestampedSnapshotImpl(SequenceNumber snapshot_seq, uint64_t ts,
...
       }
-      delete s;
       return std::make_pair(status, ret);
     } else {
       status.PermitUncheckedError();
     }
   }
 
   SnapshotImpl* snapshot =
-      snapshots_.New(s, snapshot_seq, unix_time,
```
ayulas, line 3826: the signature of the function shouldn't contains unix_time

db/snapshot_impl.h
```diff
@@ -63,6 +79,30 @@ class SnapshotList {
     list_.timestamp_ = 0;
     list_.is_write_conflict_boundary_ = false;
     count_ = 0;
+#ifdef SPEEDB_SNAP_OPTIMIZATION
+    last_snapshot_ = nullptr;
+#endif
+  }
+  SnapshotImpl* RefSnapshot(bool is_write_conflict_boundary,
+                            SequenceNumber last_pub, SystemClock* clock) {
```
ayulas, line 87: not a good name. change it to last_seq

db/snapshot_impl.h
```diff
@@ -63,6 +79,30 @@ class SnapshotList {
...
+#ifdef SPEEDB_SNAP_OPTIMIZATION
+    std::shared_ptr<SnapshotImpl> shared_snap = last_snapshot_;
+    if (shared_snap && shared_snap->GetSequenceNumber() == last_pub &&
+        shared_snap->is_write_conflict_boundary_ ==
+            is_write_conflict_boundary) {
+      SnapshotImpl* snapshot = new SnapshotImpl;
+      int64_t unix_time;
```
ayulas, line 94: why doing double setting? set it directly to snapshot->unix_time_

db/snapshot_impl.h
```diff
@@ -81,9 +121,54 @@ class SnapshotList {
...
+    delete snap;
+    if (!deleteitem) {
+      // item has not been deleted from SnapshotList
+      return true;
+    }
+#endif
+    return false;
+  }
+
+  SnapshotImpl* New(SequenceNumber seq, SystemClock* clock,
```
ayulas, line 162: dont pass the time in the function signature

db/db_impl/db_impl.cc
```diff
@@ -3870,9 +3862,14 @@ void DBImpl::ReleaseSnapshot(const Snapshot* s) {
...
   }
   const SnapshotImpl* casted_s = reinterpret_cast<const SnapshotImpl*>(s);
+  if (snapshots_.UnRefSnapshot(casted_s)) {
+    return;
+  }
   {
     InstrumentedMutexLock l(&mutex_);
     snapshots_.Delete(casted_s);
+    std::unique_lock<std::mutex> snapshotlist_lock(snapshots_.lock);
+    casted_s = nullptr;
```
ayulas, line 3872: why you need that... you deleted the s in the Delete function. do it inside the Delete

db/snapshot_impl.h
```diff
@@ -81,9 +121,54 @@ class SnapshotList {
     return list_.prev_;
   }
 
-  SnapshotImpl* New(SnapshotImpl* s, SequenceNumber seq, uint64_t unix_time,
+#ifdef SPEEDB_SNAP_OPTIMIZATION
+  SnapshotImpl* NewSnapRef(SnapshotImpl* s, SequenceNumber seq,
```
ayulas, line 125: NewSnapRef gets the s as an input... why passing all the parameters again in the function signature. you need to create the user_snap by passing in the constructor the base s

db/snapshot_impl.h
```diff
@@ -94,15 +179,25 @@ class SnapshotList {
     s->prev_->next_ = s;
     s->next_->prev_ = s;
     count_++;
+#ifdef SPEEDB_SNAP_OPTIMIZATION
+    l.unlock();
+    return NewSnapRef(s, seq, unix_time, is_write_conflict_boundary, ts);
```
ayulas, line 184: as mention .. pass s and ts only

ofriedma:
@ayulas fixed

db/db_impl/db_impl.cc
```diff
@@ -3744,11 +3740,6 @@ SnapshotImpl* DBImpl::GetSnapshotImpl(bool is_write_conflict_boundary,
...
 DBImpl::CreateTimestampedSnapshotImpl(SequenceNumber snapshot_seq, uint64_t ts,
                                       bool lock) {
-  int64_t unix_time = 0;
-  immutable_db_options_.clock->GetCurrentTime(&unix_time)
-      .PermitUncheckedError();  // Ignore error
-  SnapshotImpl* s = new SnapshotImpl;
-
   const bool need_update_seq = (snapshot_seq != kMaxSequenceNumber);
 
   if (lock) {
```
ayulas, line 3745: do this after the check of is_snapshot_supported_. you will avoid lock unlock....like the GetSnapshotImpl

ofriedma: Done

ayulas (approved):

PR #549, opened 2023-06-11
Dirty Memory: Revert allow_stall naming from allow_delays_and_stalls

Yuval-Ariel:
During #423 , the WriteBufferManager ctor parameter allow_stall was renamed to allow_delays_and_stalls to reflect its new capabilities (delaying in addition to stalling (stopping) writes). This creates confusion and unwanted behavior so this PR reverts this change and the new functionality of #423  is controller via the old WriteBufferManager ctor parameter named allow_stall.

this PR also sets back the allow_stall default value to false to avoid changing current expected behavior.


udi-speedb:
@Yuval-Ariel - Since this is not part of the release, please add the continuous gradual delay part we have discussed and I will review when that is done. Thanks

memtable/write_buffer_manager.cc
```diff
@@ -424,6 +429,9 @@ void WriteBufferManager::UpdateControllerDelayState() {
 
   if (usage_state == UsageState::kDelay) {
     WBMSetupDelay(delay_factor);
+  } else if (usage_state == UsageState::kStop && !allow_stall_) {
```
udi-speedb, line 432: allow_stall is now used a multi-threaded contexts => needs to be atomic

memtable/write_buffer_manager.cc
```diff
@@ -520,8 +528,7 @@ void WriteBufferManager::UpdateUsageState(size_t new_memory_used,
                                           ssize_t memory_changed_size,
                                           size_t quota) {
   assert(enabled());
-  if (allow_delays_and_stalls_ == false ||
-      controllers_to_refcount_map_.empty()) {
+  if (allow_delays_ == false || controllers_to_refcount_map_.empty()) {
```
udi-speedb, line 531: This is not thread safe IMO. I suggest you have another new atomic member that is the logical and (&&) of allow_delays_ and controllers_to_refcount_map_.empty().
It will always be false if allow_delays_ == false
Update it under the controllers_map_mutex_ when controllers_to_refcount_map_ becomes empty or stops being empty.

HISTORY.md
```diff
@@ -7,9 +7,9 @@ Based on RocksDB 8.1.1
...
-setup delay requests starting from (start_delay_percent * _buffer_size) / 100 (default value is 70) (start_delay_percent is another WBM ctor parameter). Changes to the WBM's memory are tracked in WriteBufferManager::ReserveMem and FreeMem.
-Once the WBM reached its capacity, writes will be stopped using the old ShouldStall() and WBMStallWrites(). (#423)
+To use this feature, pass allow_delays = true to the ctor of WBM and the db needs to be opened with options.use_dynamic_delay = true. The WBM will setup delay requests starting from (start_delay_percent * _buffer_size) / 100 (default value is 70) (start_delay_percent is another WBM ctor parameter).
+Changes to the WBM's memory are tracked in WriteBufferManager::ReserveMem and FreeMem.
+Once the WBM reached its capacity, if allow_stall == true, writes will be stopped using the old ShouldStall() and WBMStallWrites(). If allow_stall == false and the WBM reached the capacity then writes will be delayed at a rate of WriteController::kMinWriteRate (#423)
```
udi-speedb, line 12: Shouldn't the delayed write rate when reaching capacity the calculated rate?

udi-speedb:
@Yuval-Ariel I have partially reviewed.

udi-speedb (changes requested):

memtable/write_buffer_manager.cc
Yuval-Ariel: it has always been used in a multi-threaded context. its never changed, only read from.

memtable/write_buffer_manager.cc
Yuval-Ariel: actually this is a remnant of when the insertion to controllers_to_refcount_map_ was dependant on is_dynamic_delay but in this current code, controllers_to_refcount_map_ is always non empty

HISTORY.md
Yuval-Ariel: fixed

Yuval-Ariel:
@udi-speedb  it has to be part of the release since its an integral part of #423 . so #552  will be handled elsewhere. plz review asap

Yuval-Ariel:
we've agreed that for this PR, the allow_delays_and_stalls will be reverted to allow_stall and will be used solely to control the functionality of delaying writes with the WBM. The separation to two different flags, allow_stall and allow_delay is the best approach so that users will have the most flexibility and the behavior can remain unchanged but will be handled in #552 . 
This is done in order to keep the documentation organized and not ship out an unfinished feature.

udi-speedb (approved):

Issue #550, opened 2023-06-11
db_bench: New WBM per dbs in use in every group


udi-speedb:
When using db_bench with multiple groups,  when moving to a new group, a new WBM is created, causing the dbs in use in the new group not to share the WBM with the db-s from previous groups. That should not be so. The WBM (like the block cache), should be shared among all of the db-s participating in the db_bench, independent of the group.

PR #551, opened 2023-06-11
db_bench: Create a WBM once for all db-s regardless of their use in different groups (#550)

udi-speedb:


tools/db_bench_tool.cc
```diff
@@ -105,6 +105,8 @@
 #include <io.h>  // open/close
 #endif
 
+#include "util/string_util.h"
+
```
Yuval-Ariel, line 109: already included

tools/db_bench_tool.cc
```diff
@@ -3652,9 +3652,13 @@ class Benchmark {
     std::unique_ptr<ExpiredTimeFilter> filter;
     while (std::getline(benchmark_stream, name, ',')) {
       if (open_options_.write_buffer_manager) {
-        fprintf(stderr, "\nBEFORE Benchmark (%s): %lu OF %lu\n\n", name.c_str(),
-                open_options_.write_buffer_manager->memory_usage(),
-                open_options_.write_buffer_manager->buffer_size());
+        fprintf(stderr, "\nBEFORE Benchmark (%s): %s OF %s\n\n", name.c_str(),
```
Yuval-Ariel, line 3655: while you're here, can you also add info to the printing that its the WBM's memory you're reporting?

Yuval-Ariel: right now it looks like this:
------------------------------------------------

BEFORE Benchmark (fillseq): 6144 OF 110000000

udi-speedb: Done. And now to the Approve button please

Yuval-Ariel (approved):
:)

Yuval-Ariel (approved):

PR #561, opened 2023-06-13
560: Tombstone test fixes as a result of not clear local variable

ayulas:


ayulas:
@yuval pls review . for some reason the unit test passed on the CI but not on the mac. now it is passed on the mac as well
it must inserted to the version... no need performance test to it

ayulas:
toda!!!

Issue #577, opened 2023-06-25
Block cache stats are printed to the log on the same line as Write Stall (count): log trace


udi-speedb:
**Describe the bug**
After the rebase on RocksDB 8.1.1, this is what the log shows:
Write Stall (count): cf-l0-file-count-limit-delays-with-ongoing-compaction: 0, cf-l0-file-count-limit-stops-with-ongoing-compaction: 0, l0-file-count-limit-delays: 0, l0-file-count-limit-stops: 0, memtable-limit-delays: 0, memtable-limit-stops: 0, pending-compaction-bytes-delays: 0, pending-compaction-bytes-stops: 0, total-delays: 0, total-stops: 0, Block cache LRUCache@0x55adc40e3060#22075 capacity: 2.05 GB usage: 0.09 KB table_size: 1024 occupancy: 88 collections: 1 last_copies: 
0 last_secs: 6.9e-05 secs_since: 0

**Expected behavior**
The block cache information should be printed on a separate line
Write Stall (count): cf-l0-file-count-limit-delays-with-ongoing-compaction: 0, cf-l0-file-count-limit-stops-with-ongoing-compaction: 0, l0-file-count-limit-delays: 0, l0-file-count-limit-stops: 0, memtable-limit-delays: 0, memtable-limit-stops: 0, pending-compaction-bytes-delays: 0, pending-compaction-bytes-stops: 0, total-delays: 0, total-stops: 0

Block cache LRUCache@0x55adc40e3060#22075 capacity: 2.05 GB usage: 0.09 KB table_size: 1024 occupancy: 88 collections: 1 last_copies: 0 last_secs: 6.9e-05 secs_since: 0

PR #578, opened 2023-06-25
Fix log trace of Block Cache (Entry) Stats dump (#577)

udi-speedb:


Yuval-Ariel (approved):

db/internal_stats.cc
```diff
@@ -739,7 +740,7 @@ std::string InternalStats::CacheEntryRoleStats::ToString(
           << (100.0 * total_charges[i] / cache_capacity) << "%)";
     }
   }
-  str << "\n";
+  str << '\n';
```
Yuval-Ariel, line 743: why is this necessary?

udi-speedb: Without it the next dump entity comes without any separating spaces. Makes it harder to read for a human.

Yuval-Ariel: i remind you that we've discussed that this change should be removed

udi-speedb: Forced pushed the update

PR #589, opened 2023-07-02
unit tests: Fix GlobalAndWBMSetupDelay

Yuval-Ariel:
RUN      ] GlobalWriteControllerTest.GlobalAndWBMSetupDelay
db/global_write_controller_test.cc:391: Failure
Value of: options.write_buffer_manager->memory_usage() < 5_kb
Actual: false
Expected: true
FAILED  ] [GlobalWriteControllerTest.GlobalAndWBMSetupDelay (44 ms)

The memory is freed only when the detor of the version is called in `job_context.Clean();` under `DBImpl::BackgroundCallFlush` .

The problem was that calling Flush with `FlushOptions.wait = true`, doesn't wait for the memory to be free which also happens during the flush process.
So the fix is adding a sync point which delays the assertion of checking the WBM memory usage to after the flush cleared the memtable memory.

### DIFF
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -5,6 +5,9 @@
 ### Enhancements
 * db_bench: add estimate-table-readers-mem benchmark which prints these stats.
 
+### Bug Fixes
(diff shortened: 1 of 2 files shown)


ofriedma (approved):
LGTM

PR #612, opened 2023-07-24
548: Support Spdb memtable in Java and C

ayulas:


RoyBenMoshe (approved):

mrambacher (commented):
This is not the preferred way of adding code in the plugin directory into Java.  This now makes Java dependent on compiling with the Speedb features enabled and requires the code to be managed/merged between RocksDB and Speedb versions.

If the goal is to minimize Speedb changes in the base RocksDB code, then this code should be separated out into its own files/directory under the speedb plugin and added to Java accordingly.

HISTORY.md
```diff
@@ -64,7 +64,7 @@ Also switch to waiting a sec on the CV each time. This is required since a bg er
 * db_bench: Create a WBM once for all db-s regardless of their use in different groups (#550)
 * Tompstone unit test faiure (#560)
 * build: Remove unused variables in unit tests (#581)
-
+* support hash_spdb in java (#548)
```
Yuval-Ariel, line 67: this needs to 2.5 is already released. you need to add changes under the "Unreleased" section

Yuval-Ariel:
it seems that this PR also make the speedb memtable not a plugin, right?
if so then either update the description or better yet, divide into 2 separate PRs

mrambacher:
For Java, this could have been done via the CreateFromString in the same manner that Filters were exposed.  This would have eliminated the need for all of the custom classes and changing the code to a plugin.  Is there a reason this approach was not chosen?

Something similar could have been implemented in C, but I am not aware of any exposures of CreateFromString in that language as of yet.

PR #619, opened 2023-08-01
WIP: Allow Serialization of only options that have changed

mrambacher:
This PR does a few things:
1) Changes the Configurable::SerializeOptions method to return a map of options.  This map is equivalent to the input map for the ConfigureOptions method.  By returning a map, it will be possible in the future to add different output representations (such as DUMP, JSON, Properties) formats to the string representations of the configurable objects.

2) Add the ability to only serialize options that have changed.  If the "ConfigOptions::compare_to" value is set, only option values that do not match those in the "compare_to" will be part of the serialization.  There are a couple places where one could envision using this parameter.  One use case is to only write to the log file options that are not the default (e.g. only print ColumnFamilyOptions that are not the default value) or that do not match some default (e.g. only print Column Family Options that do not match those for the default column family).  This "printing of only changed options" could be useful in printing less information in the LOG and OPTIONS file, thereby speeding them up and making the contents more concise.

### DIFF
--- a/db/compaction/compaction_service_job.cc
+++ b/db/compaction/compaction_service_job.cc
@@ -693,8 +693,8 @@ static std::unordered_map<std::string, OptionTypeInfo> cs_result_type_info = {
         const auto status_obj = static_cast<const Status*>(addr);
         StatusSerializationAdapter adapter(*status_obj);
         std::string result;
-        Status s = OptionTypeInfo::SerializeType(opts, status_adapter_type_info,
-                                                 &adapter, &result);
+        Status s = OptionTypeInfo::TypeToString(
+            opts, "", status_adapter_type_info, &adapter, &result);
         *value = "{" + result + "}";
         return s;
       },
@@ -770,7 +770,13 @@ Status CompactionServiceInput::Write(std::string* output) {
   output->append(buf, sizeof(BinaryFormatVersion));
   ConfigOptions cf;
   cf.invoke_prepare_options = false;
-  return OptionTypeInfo::SerializeType(cf, cs_input_type_info, this, output);
+  std::unordered_map<std::string, std::string> options;
+  Status s =
+      OptionTypeInfo::SerializeType(cf, cs_input_type_info, this, &options);
+  if (s.ok()) {
+    output->append(cf.ToString("", options) + cf.delimiter);
+  }
+  return s;
 }
 
 Status CompactionServiceResult::Read(const std::string& data_str,
@@ -799,7 +805,13 @@ Status CompactionServiceResult::Write(std::string* output) {
   output->append(buf, sizeof(BinaryFormatVersion));
   ConfigOptions cf;
   cf.invoke_prepare_options = false;
-  return OptionTypeInfo::SerializeType(cf, cs_result_type_info, this, output);
+  std::unordered_map<std::string, std::string> options;
+  Status s =
+      OptionTypeInfo::SerializeType(cf, cs_result_type_info, this, &options);
+  if (s.ok()) {
+    output->append(cf.ToString("", options) + cf.delimiter);
+  }
+  return s;
 }
 
 #ifndef NDEBUG
--- a/include/rocksdb/configurable.h
+++ b/include/rocksdb/configurable.h
@@ -343,8 +343,9 @@ class Configurable {
                                std::string* bad_name) const;
   // Internal method to serialize options (ToString)
   // Classes may override this value to change its behavior.
-  virtual std::string SerializeOptions(const ConfigOptions& config_options,
-                                       const std::string& header) const;
+  virtual Status SerializeOptions(
+      const ConfigOptions& config_options,
+      std::unordered_map<std::string, std::string>* options) const;
 
   //  Given a name (e.g. rocksdb.my.type.opt), returns the short name (opt)
   virtual std::string GetOptionName(const std::string& long_name) const;
--- a/include/rocksdb/convenience.h
+++ b/include/rocksdb/convenience.h
@@ -93,13 +93,15 @@ struct ConfigOptions {
   // The object registry to use for this options
   std::shared_ptr<ObjectRegistry> registry;
 
+  // If set, only changes from this reference version will be serialized.
+  Configurable* compare_to = nullptr;
+
   bool IsShallow() const { return depth == Depth::kDepthShallow; }
   bool IsDetailed() const { return depth == Depth::kDepthDetailed; }
 
   bool IsCheckDisabled() const {
     return sanity_level == SanityLevel::kSanityLevelNone;
   }
-
   bool IsCheckEnabled(SanityLevel level) const {
     return (level > SanityLevel::kSanityLevelNone && level <= sanity_level);
   }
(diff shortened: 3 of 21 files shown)


udi-speedb:
@mrambacher - Please update when you have completed updating the branch. I will review only once the branch has all of the contents you intend to merge. Thanks

mrambacher:
A couple of points:
1. I do not know if it is better to return/use a map or a vector of string pairs.  Vectors may be slightly easier to work with in some cases and more efficient.  They would also preserve the order if that is important (whereas map order is more random).  The Configure methods currently use maps (but perhaps an internal wrapper method should be created to use the same type)
2. I was going to create something like a typedef/using for the map but could not decide on a good name.  Feel free to suggest something as the current type is pretty long to type...

udi-speedb:
@mrambacher - Do you still have content to push to the PR?

mrambacher:
> @mrambacher - Do you still have content to push to the PR?

@udi-speedb I just made an update to use the Properties type.  I am done now pending comments.

db/compaction/compaction_service_job.cc
```diff
@@ -693,8 +693,8 @@ static std::unordered_map<std::string, OptionTypeInfo> cs_result_type_info = {
         const auto status_obj = static_cast<const Status*>(addr);
         StatusSerializationAdapter adapter(*status_obj);
         std::string result;
-        Status s = OptionTypeInfo::SerializeType(opts, status_adapter_type_info,
-                                                 &adapter, &result);
+        Status s = OptionTypeInfo::TypeToString(
```
udi-speedb, line 696: please replace "" with "" /* prefix */ for clarity

include/rocksdb/utilities/options_type.h
```diff
@@ -855,9 +855,13 @@ class OptionTypeInfo {
   // Returns the serialized form in result.
   // Returns OK on success or non-OK if some option could not be serialized.
   static Status SerializeType(
-      const ConfigOptions& config_options,
+      const ConfigOptions& config_options, const std::string& prefix,
```
udi-speedb, line 858: What is the difference between SerializeType() and TypeToString()?

include/rocksdb/utilities/options_type.h
```diff
@@ -855,9 +855,13 @@ class OptionTypeInfo {
...
+      const std::unordered_map<std::string, OptionTypeInfo>& type_map,
+      const void* opt_addr, Properties* props);
+  static Status TypeToString(
+      const ConfigOptions& config_options, const std::string& opt_name,
```
udi-speedb, line 862: is the opt_name parameter the prefix?

include/rocksdb/utilities/options_type.h
```diff
@@ -1156,33 +1169,22 @@ Status SerializeVector(const ConfigOptions& config_options,
                        const OptionTypeInfo& elem_info, char separator,
                        const std::string& name, const std::vector<T>& vec,
                        std::string* value) {
```
udi-speedb, line 1171: How are these changes related to this PR?

include/rocksdb/configurable.h
```diff
@@ -25,6 +25,8 @@ struct ColumnFamilyOptions;
 struct ConfigOptions;
 struct DBOptions;
 
+using Properties = std::unordered_map<std::string, std::string>;
```
udi-speedb, line 28: This is duplicated in convenience.h.
There should be one such definition in a common location
And again, should be a some narrower scope / namespace

udi-speedb:
I may be missing something, but how is the addition of the prefix parameter part of this PR?

options/cf_options.cc
```diff
@@ -1157,7 +1157,8 @@ Status GetStringFromMutableCFOptions(const ConfigOptions& config_options,
                                      std::string* opt_string) {
   assert(opt_string);
   opt_string->clear();
-  return OptionTypeInfo::SerializeType(
-      config_options, cf_mutable_options_type_info, &mutable_opts, opt_string);
+  return OptionTypeInfo::TypeToString(config_options, "",
```
udi-speedb, line 1160: "" -> "" /* prefix */

options/cf_options.cc
```diff
@@ -1157,7 +1157,8 @@ Status GetStringFromMutableCFOptions(const ConfigOptions& config_options,
...
+                                      cf_mutable_options_type_info,
```
udi-speedb, line 1161: Why have you moved to TypeToString()?

include/rocksdb/convenience.h
```diff
@@ -103,6 +105,12 @@ struct ConfigOptions {
   bool IsCheckEnabled(SanityLevel level) const {
     return (level > SanityLevel::kSanityLevelNone && level <= sanity_level);
   }
+  // Converts the properties to a single string representation
```
udi-speedb, line 108: How are these 2 methods related to this PR?

udi-speedb (changes requested):

include/rocksdb/convenience.h
```diff
@@ -23,6 +23,8 @@ struct ColumnFamilyOptions;
 struct DBOptions;
 struct Options;
 
+using Properties = std::unordered_map<std::string, std::string>;
```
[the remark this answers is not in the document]
mrambacher: Renamed to OptionProperties

include/rocksdb/convenience.h
```diff
@@ -103,6 +105,12 @@ struct ConfigOptions {
   bool IsCheckEnabled(SanityLevel level) const {
     return (level > SanityLevel::kSanityLevelNone && level <= sanity_level);
   }
+  // Converts the properties to a single string representation
+  std::string ToString(const std::string& prefix,
```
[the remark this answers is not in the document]
mrambacher: The prefix parameter was there previously (called header) and is needed by some of the formats (like Log format)

include/rocksdb/convenience.h
mrambacher: The ToString methods are used to convert the UserProperties to their string representation.  Prior, the code was "hard-coded" in the various places

db/db_basic_test.cc
```diff
@@ -2197,6 +2197,8 @@ TEST_F(DBBasicTest, DBSetThreadAffinity) {
     CPU_ZERO(&cpuset);
     CPU_SET(0, &cpuset);
     pthread_setaffinity_np(thr, sizeof(cpu_set_t), &cpuset);
+#else
+    (void)thr;
```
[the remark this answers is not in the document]
mrambacher: This is here because of a general build issue on Macs.  I have fixed this in many PRs now.  Hopefully one will take...

mrambacher:
> I may be missing something, but how is the addition of the prefix parameter part of this PR?

The prefix parameter to SerializeOptions was there before -- the variable was called "header"

options/cf_options.cc
mrambacher: SerializeType now returns a OptionsProperties set.  TypeToString converts the Properties to a string.  There were multiple places that needed this functionality, so I put it under a single API/method.

udi-speedb (approved):

PR #632, opened 2023-08-10
Proactive Flushes: Decrement number of flushes to init on non-wbm-initiated flush end

udi-speedb:


memtable/write_buffer_manager.cc
```diff
@@ -731,6 +739,12 @@ void WriteBufferManager::FlushStarted(bool wbm_initiated) {
   flushes_mu_->Lock();
 
   ++num_running_flushes_;
+  // Any number of non-wbm-initiated flushes may be initiated, so, we must not
+  // underflow num_flushes_to_initiate_
```
Yuval-Ariel, line 743: what does underflow mean?

udi-speedb: Decreasing the counter below 0 (wraps around to max)

Yuval-Ariel (approved):

PR #655, opened 2023-08-29
653: Support hash spdb as part of enable speedb features 

ayulas:


RoyBenMoshe (approved):

memtable/hash_spdb_rep.cc
```diff
@@ -552,22 +545,12 @@ static std::unordered_map<std::string, OptionTypeInfo> hash_spdb_factory_info =
          {offsetof(struct HashSpdbRepOptions, hash_bucket_count),
           OptionType::kSizeT, OptionVerificationType::kNormal,
           OptionTypeFlags::kNone}},
-        {"use_seek_parallel_threshold",
-         {offsetof(struct HashSpdbRepOptions, use_seek_parallel_threshold),
-          OptionType::kBoolean, OptionVerificationType::kNormal,
-          OptionTypeFlags::kNone}},
```
mrambacher, line 558: If anyone has an Options file with this written to it, removing this will break backwards compatibility.  The standard practice is to set the offset to 0 and use the deprecated flag.  This way options files which have this value will continue to work but the value will be ignored.

Yuval-Ariel:
whats the status here?

Yuval-Ariel (approved):

Issue #659, opened 2023-09-01
MacOS does not support thread affinity and fails to build


mrambacher:
g++  -fPIC -DROCKSDB_DLL -DROCKSDB_USE_RTTI    -g -W -Wextra -Wall -Wsign-compare -Wshadow -Wunused-parameter -Werror -I. -I./include -std=c++17  -faligned-new -DHAVE_ALIGNED_NEW -DROCKSDB_PLATFORM_POSIX -DROCKSDB_LIB_IO_POSIX  -DOS_MACOSX -DSNAPPY -DGFLAGS=1 -DZLIB -DBZIP2 -DLZ4 -DZSTD -DROCKSDB_BACKTRACE -Wshorten-64-to-32 -DHAVE_UINT128_EXTENSION -DHAVE_BARRIERFSYNC -DHAVE_FULLFSYNC  -isystem third-party/gtest-1.8.1/fused-src  -DROCKSDB_ASSERT_STATUS_CHECKED -Woverloaded-virtual -Wnon-virtual-dtor -Wno-missing-field-initializers -Wno-invalid-offsetof -c db/db_basic_test.cc -o db/db_basic_test.o
db/db_basic_test.cc:2180:5: error: unknown type name 'cpu_set_t'
    cpu_set_t cpuset;
    ^
db/db_basic_test.cc:2183:40: error: use of undeclared identifier 'cpu_set_t'; did you mean 'cpuset'?
    pthread_setaffinity_np(thr, sizeof(cpu_set_t), &cpuset);
                                       ^~~~~~~~~
                                       cpuset
db/db_basic_test.cc:2180:15: note: 'cpuset' declared here
    cpu_set_t cpuset;
              ^
2 errors generated.
make: *** [db/db_basic_test.o] Error 1

speedb % cc -v
Apple clang version 14.0.3 (clang-1403.0.22.14.1)
Target: arm64-apple-darwin22.6.0
Thread model: posix
InstalledDir: /Library/Developer/CommandLineTools/usr/bin

PR #661, opened 2023-09-03
Add MacOS support for thread affinity tests

ofriedma:
fixes: https://github.com/speedb-io/speedb/issues/659


ofriedma:
@mrambacher the callback will be empty for platforms that are not windows or linux.

db/db_basic_test.cc
```diff
@@ -2175,7 +2175,8 @@ TEST_F(DBBasicTest, DBSetThreadAffinity) {
 #if defined(OS_WIN)
 #include "winbase.h"
     SetThreadAffinityMask(thr, 0);
-#else
```
mrambacher, line 2178: why not elif (OS_LINUX) ?

db/db_basic_test.cc
```diff
@@ -2175,7 +2175,8 @@ TEST_F(DBBasicTest, DBSetThreadAffinity) {
 #if defined(OS_WIN)
 #include "winbase.h"
```
mrambacher, line 2176: These headers (pthread, winbase) should be included at the top and in <> rather than "" since they are system header files.

mrambacher (approved):
Minor comments.  Thanks for the quick fix

mrambacher (approved):
If it passes make format (I thought system includes were supposed to be first), good enough for me.

PR #669, opened 2023-09-07
build test - Windows, Mac and Ubuntu

maxb-io:


udi-speedb:
@maxb-io - What is the status of this PR? Who should do what about this?

PR #670, opened 2023-09-07
Export GetCompactionReasonString/GetFlushReasonString by moving them into listener.h

git-hulk:
Currently, rocksdb users would use the event listener to catch the compaction/flush event and log them if any. But now the reason is an integer type instead of a human-readable string, so we would like to convert them into a human-readable string.

Refer: https://github.com/facebook/rocksdb/pull/11778

### DIFF
--- a/include/rocksdb/listener.h
+++ b/include/rocksdb/listener.h
@@ -161,6 +161,8 @@ enum class CompactionReason : int {
   kNumOfReasons,
 };
 
+const char* GetCompactionReasonString(CompactionReason compaction_reason);
+
 enum class FlushReason : int {
   kOthers = 0x00,
   kGetLiveFiles = 0x01,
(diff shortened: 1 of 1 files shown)


git-hulk:
Not sure if it's the right way to submit the PR which was merged in rocksdb. Can help to correct me if I was wrong?

udi-speedb:
We are rebasing on RocksDB from time to time, so that is the way we usually sync with RocksDB.
If you are using Speedb and need it urgently, I can check internally if we may accept the PR individually.

git-hulk:
> We are rebasing on RocksDB from time to time, so that is the way we usually sync with RocksDB. If you are using Speedb and need it urgently, I can check internally if we may accept the PR individually.

Hi @udi-speedb, thanks for your reply. 

We're in no hurry, let's close this PR and wait for rebasing from upstream.

Issue #637, opened 2023-08-15
deadlock in write _thread linkone function during stall


ayulas:
deadlock stall

ayulas:
the thread which begin the stall:

 futex_wait_cancelable (private=<optimized out>, expected=0,
    futex_word=0x7f632d86dec8) at ../sysdeps/nptl/futex-internal.h:183
#1  __pthread_cond_wait_common (abstime=0x0, clockid=0, mutex=0x7f632d86de78,
    cond=0x7f632d86dea0) at pthread_cond_wait.c:508
#2  __pthread_cond_wait (cond=0x7f632d86dea0, mutex=0x7f632d86de78)
    at pthread_cond_wait.c:647
#3  0x00005632819bffe1 in rocksdb::port::CondVar::Wait (this=<optimized out>)
    at port/port_posix.cc:119
#4  0x00005632817dfea1 in rocksdb::DBImpl::WBMStallInterface::Block (
    this=0x7f632d86de70) at ./db/db_impl/db_impl.h:1249
#5  rocksdb::DBImpl::WriteBufferManagerStallWrites (this=0x7f632d87ec00)
    at db/db_impl/db_impl_write.cc:1932
#6  0x00005632817e54fa in rocksdb::DBImpl::PreprocessWrite (
    this=<optimized out>, write_options=..., log_context=0x7f62c85d5660,
    write_context=<optimized out>) at db/db_impl/db_impl_write.cc:1228
#7  0x00005632817e99de in rocksdb::DBImpl::WriteImpl (this=0x7f632d87ec00,
    write_options=..., my_batch=<optimized out>, callback=<optimized out>,
    log_used=<optimized out>, log_ref=<optimized out>,
    disable_memtable=<optimized out>, seq_used=<optimized out>,
    batch_cnt=<optimized out>, pre_release_callback=<optimized out>,
    post_memtable_callback=<optimized out>) at db/db_impl/db_impl_write.cc:391
#8  0x00005632817ead44 in rocksdb::DBImpl::Write (this=0x7f632d87ec00,
    write_options=..., my_batch=0x7f62c85d5da0)
    at db/db_impl/db_impl_write.cc:146


the call WriteBufferManagerStallWrites is under db mutex which leads to a potential dead lock incase  WriteThread::LinkOne getting into stall wait.

ayulas:
there is a race that wbm stall doesnt trigger the flush so the mem is not free. therefor the stall will never ended and all the writes are stucked

[...]

Issue #653, opened 2023-08-29
Support hash spdb as part of enable speedb features


ayulas:
Support hash spdb as part of enable speedb features

Guyme:
@ayulas - please rebase and re-run CI.
Once done please notify @Yuval-Ariel to merge.

PR #683, opened 2023-09-21
637 stall deadlock

ayulas:
fixes : #637


udi-speedb:
@ayulas I very well may be totally wrong, but is this really the code to solve the deadlock in linkone?
I am asking only since I have briefly looked at the changes and it has made me wonder.

Guyme:
@ayulas - status please.
also please verify that https://github.com/speedb-io/speedb/pull/683 is indeed the code for this deadlock (the content seams to be something else - history updates referes https://github.com/speedb-io/speedb/issues/653)

Issue #687, opened 2023-09-27
db_bench usability of the pinning policy parameter 


erez-speedb:
The current parameter is undocumented and complicated:
-pinning_policy=""id=speedb_scoped_pinning_policy;capacity=$(($cache_size * 80 /100))"

I suggest something simple like:
-pinning_policy=speedb_scoped_pinning_policy;
-pinnig_capacity= % from block cache

This should also imply -cache_index_and_filter_blocks

erez-speedb:
@udi-speedb  @hilikspdb  based on what were those numbers chosen?

udi-speedb:
I am going to do the following:

1. Cache size is set independently by the user (or implicitly). db_bench should do the calculations on its own rather than expect the user to provide a number.
2. I will take into account the WBM if enabled and configured to cost to cache. So, the pinning policy will be configured to use capacity = 80% of (cache-size - WBM (if applicable) )
3. If the user specifies a pinning policy and configured cache_index_and_filter_blocks=false, db_bench will report the mis-configuration and exit.
4. The default pinning policy is to use a policy that should result in the same behavior as the one in RocksDB or that was in effect before static pinning feature was merged (=> pinning based on the multiple existing pinning options). The alternative is to use our own Scoped pinning policy that pins based on the thresholds, ignoring the pinning options. Although the way it is implemented in db_bench is generic, and support any pinning policy, it is inconvenient (that's how this issue came to be). So, I will have the following db_bench flags to control pinning:
- scoped_pinning_policy: Boolean. False by default => Default pinning policy. When true, scoped pinning policy.

- scoped_pinning_last_level_with_data_percent: Sets the bottom threshold (applicable only when scoped_pinning_policy=true). Default: Same as the defaults set in the code (currently 10% but will be synchronized with whatever is set in the code (soon will be 70%)

- scoped_pinning_mid_percent: Sets the middle (non-level-0) threshold (applicable only when scoped_pinning_policy=true).  Default: Same as the defaults set in the code (currently 10% but will be synchronized with whatever is set in the code (soon will be 90%)

scoped_pinning_mid_percent must be >= scoped_pinning_last_level_with_data_percent

udi-speedb:
> @udi-speedb @hilikspdb based on what were those numbers chosen?

Which numbers? Should this question be asked in #689 ?

mrambacher:
[...]

PR #699, opened 2023-10-03
Pinning policy was changed to scoped pinning in enableSpeedbFeaturesCF

RoyBenMoshe:


RoyBenMoshe:
This PR should be merged only after PR (#653 ) is merged

options/options.cc
```diff
@@ -627,25 +628,15 @@ ColumnFamilyOptions* ColumnFamilyOptions::EnableSpeedbFeaturesCF(
...
-    cache_usage_options.options_overrides.insert(
-        {CacheEntryRole::kBlockBasedTableReader, role_options});
-    cache_usage_options.options_overrides.insert(
-        {CacheEntryRole::kCompressionDictionaryBuildingBuffer, role_options});
-    cache_usage_options.options_overrides.insert(
-        {CacheEntryRole::kFileMetadata, role_options});
+    std::ostringstream oss;
+    oss << "id=speedb_scoped_pinning_policy; capacity="
+        << 0.8 * shared_options.GetTotalRamSizeBytes()
+        /* << "; bottom_percent=10; mid_percent=80"*/;
```
udi-speedb, line 635: The capacity for pinning should be 0.8 of the capacity we intend to allocate to pinning. That should be:
0.8 X (Total cache size - write buffer's manager size)
Total cache size should be using the actual configured size of the cache and not the total ram size the user has specified for the entire system IMO.
The above assumes we are using a WBM, and that it is configured to cost to cache.

udi-speedb:
@RoyBenMoshe, @ayulas - I know I am not the reviewer, however, since I am working on static pinning, and making changes that should be in sync (e.g., #687), I briefly looked at the code and made the comment above.

udi-speedb:
Closing as this was replaced by https://github.com/speedb-io/speedb/pull/715

PR #710, opened 2023-10-11
706 move ci tests to gh shared runners

maxb-io:
switch self-hosted github runners with github shared runners


speedbadmin (approved):

maxb-io:
Leaving replaced commands in comment for some time

PR #711, opened 2023-10-12
697 unit tests some tests are stuck

Yuval-Ariel:


udi-speedb:
@Yuval-Ariel - Please update HISTORY.md (do we need to add the our licencse text at the top here?)

udi-speedb (approved):

PR #720, opened 2023-10-16
db bench: usability of the pinning policy parameter

udi-speedb:


tools/db_bench_tool.cc
```diff
@@ -818,7 +819,27 @@ DEFINE_bool(
     " Note `cache_index_and_filter_blocks` must be true for this option to have"
     " any effect.");
 
-DEFINE_string(pinning_policy, "", "URI for registry TablePinningPolicy");
+DEFINE_bool(scoped_pinning_policy, false, "Use Speedb's Scoped Pinning Policy");
```
mrambacher, line 822: Can we please leave this as a string?  Boolean parameters make it much harder to do validation and more alternatives.

mrambacher: The string could be compared to the kClassName or the like...

tools/db_bench_tool.cc
```diff
@@ -9280,6 +9314,46 @@ void ValidateMetadataCacheOptions() {
...
+        FLAGS_scoped_pinning_mid_percent) {
+      ErrorExit(
+          "--scoped_pinning_last_level_with_data_percent must be <= "
+          "--scoped_pinning_mid_percent must be between 0 and 100");
+    }
+
+    if (FLAGS_cost_write_buffer_to_cache) {
+      if (FLAGS_db_write_buffer_size > FLAGS_cache_size) {
+        ErrorExit("--cache_size must be >= --db_write_buffer_size");
+      }
```
mrambacher, line 9352: What does this have to do with the pinning policy?

tools/db_bench_tool.cc
```diff
@@ -9597,6 +9671,7 @@ int db_bench_tool_run_group(int group_num, int num_groups, int argc,
   }
 
   ValidateMetadataCacheOptions();
+  ValidatePinningRelatedOptions();
```
mrambacher, line 9674: Aren't these two mutually exclusive?  The MetadataCacheOptions are not used if PinningRelatedOptions are used

tools/db_bench_tool.cc
udi-speedb: Indirectly. These flags are used later in the calculation of the pinnning capacity.

tools/db_bench_tool.cc
udi-speedb: The whole point was to make it easy for users such as @erez-speedb to command db_bench to use scoped pinning policy. A boolean achieves just that. The whole point of this PR is usability and making the lives of the user easy.

And what validations are you referring to in this context?

mrambacher: If we add a third pinning policy and the value is a boolean, then you can only set one of the booleans.  A string is simpler and more expandable/adaptable.  Look at the code for handling different Envs to see an example.

I believe the problem before was the string that was required was too complex.  This code simplifies the string (and provides validation) by separating the fields that were in one string into multiple parameters.  

And is it really that much harder to say --pinning_policy=scoped than --scoped_pinning_policy?

tools/db_bench_tool.cc
mrambacher: Are the MetadataCacheOptions validation a problem with db_bench or the code in general?

I guess leave it for now.  Not worth the hassle.

tools/db_bench_tool.cc
udi-speedb: I have reverted it to a string but the values are NOT the Class names (which are long and hard to remember. @erez-speedb complained about the length of the command line as it is).
The values (at the moment are: "default" and "scoped")

mrambacher: I strongly suggest making the values class names and changing the class name if you do not like it.  Using name/nickname of "scoped" seems very reasonable to me.  The MergeOperators with names like "MergeSortOperator" and "sortlist" are great examples of names and nicknames that make sense

db_stress_tool/db_stress_gflags.cc
```diff
@@ -890,7 +891,12 @@ DEFINE_string(fs_uri, "",
               " with --env_uri."
               " Creates a default environment with the specified filesystem.");
 
-DEFINE_string(pinning_policy, "", "URI for registry TablePinningPolicy");
+DEFINE_string(pinning_policy,
+              ROCKSDB_NAMESPACE::DefaultPinningPolicy::kNickName(),
+              "The pinning policy to use. "
+              "The options are: "
+              "'DefaultPinning': Default RocksDB's pinning polcy. "
+              "'ScopedPinning': Speedb's Scoped pinning policy.");
```
mrambacher, line 899: Why explicitly state what the options are?  If more were added via a plugin, this would be wrong.  It also makes it hard to maintain.

udi-speedb: Obviously, however, the top priority here IMO is to make it easy for the users (and seeing the strings to use does just that). That is a purpose of a help string.
Adding a new policy is a rare event. Such a policy would probably have its own set of configuration flags that should be added anyway to db_bench, so, I believe that the developer adding the policy would update the help string / default if necessary.

plugin/speedb/pinning_policy/scoped_pinning_policy.h
```diff
@@ -22,24 +22,27 @@ namespace ROCKSDB_NAMESPACE {
...
 
-// A table policy that limits the size of the data to be pinned
 //
 class ScopedPinningPolicy : public RecordingPinningPolicy {
  public:
   ScopedPinningPolicy();
   ScopedPinningPolicy(const ScopedPinningOptions& options);
 
   static const char* kClassName() { return "speedb_scoped_pinning_policy"; }
-  static const char* kNickName() { return "speedb.ScopedPinningPolicy"; }
```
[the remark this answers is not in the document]
udi-speedb: Done

udi-speedb:
@mrambacher - I have pushed another commit. Hopefully the last one. Please see if there is anything that you really think is important enough. Otherwise, please approve. Thanks

Yuval-Ariel:
@udi-speedb note that theres a conflict

udi-speedb:
> @udi-speedb note that theres a conflict
@Yuval-Ariel - Thanks.
I am waiting for @mrambacher to complete the review and then I will resolve the conflicts

udi-speedb:
> @udi-speedb note that theres a conflict

Thanks. Resolved

db_stress_tool/db_stress_gflags.cc
```diff
@@ -890,7 +891,12 @@ DEFINE_string(fs_uri, "",
               " with --env_uri."
               " Creates a default environment with the specified filesystem.");
 
-DEFINE_string(pinning_policy, "", "URI for registry TablePinningPolicy");
+DEFINE_string(pinning_policy,
+              ROCKSDB_NAMESPACE::DefaultPinningPolicy::kNickName(),
```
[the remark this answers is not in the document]
mrambacher: Setting this to a legal value rather than empty makes it much harder to tell if it is updated.  If it is empty, then the system will use default.  If not, it should use the one that is set.

mrambacher (approved):
thanks

PR #742, opened 2023-11-03
Enable speedb features: Constrain the i/f of SharedOptions (make immutable)

udi-speedb:


examples/enable_speedb_features_example.cc
```diff
@@ -55,11 +56,17 @@ int main() {
   // as listed in the definition of SpeedbSharedOptiopns in options.h
   op1.create_if_missing = true;
   op1.compression = rocksdb::kNoCompression;
+  // NOT Having a prefix-extractor (the deafult) will result in the
```
ofriedma, line 59: having

options/options.cc
```diff
@@ -551,33 +551,32 @@ Options* Options::OldDefaults(int rocksdb_major_version,
 Options* Options::EnableSpeedbFeatures(SharedOptions& shared_options) {
   EnableSpeedbFeaturesDB(shared_options);
   EnableSpeedbFeaturesCF(shared_options);
+  if (memtable_factory->IsInsertConcurrentlySupported() == false) {
+    allow_concurrent_memtable_write = false;
```
ofriedma, line 555: Assert for a configuration conflict

ofriedma (changes requested):

examples/enable_speedb_features_example.cc
udi-speedb: Done

ofriedma (approved):

udi-speedb:
@ofriedma - Please resolve the comments. Thanks

options/options.cc
ofriedma: where is the assert? @udi-speedb

PR #759, opened 2023-11-08
Proactive Flushes: Fix a race in the ShouldInitiateAnotherFlushMemOnly that may cause the method to return an incorrect answer

udi-speedb:


ofriedma (approved):
It solves potential int overflow, LGTM

Issue #743, opened 2023-11-03
db_bench: Sync applicable db_bench's properties with enable speedb features


udi-speedb:
When activating  enable speedb features in db_bench, we are creating and overriding Options' properties.
db_bench has its own. 
For example, db_bench has its own cache handling that seems not to be fully synchronised with the SharedOptions's own cache entity:

1. no_block_cache is set according to the db_bench's cache_ member.
2. db_bench creates a SimCache as its cache if simcache_size >= 0.

We should review all the overrides that enables speedb features does and make sure they are fully synchronized with db_bench.

ofriedma:
https://github.com/speedb-io/speedb/pull/765

ofriedma:
@erez-speedb moved to performance in order to check if it doesn't fail for the performance tests, it doesn't affect the performance of the tool or the db itself.

erez-speedb:
Passed perf test on hash 06fcbd35

PR #765, opened 2023-11-12
fix conflicts between db_bench flags and enable speedb features flag

ofriedma:
fixes: https://github.com/speedb-io/speedb/issues/743

### DIFF
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -11,9 +11,11 @@ Based on RocksDB 8.6.7
 * Added a kUseBaseAddress flag and GetBaseOffset flag to OptionTypeInfo.  If this flag is set and a function is used for processing options, the function is passed the base address of the struct rather than the specific field (#397)
 
 ### Bug Fixes
-* Stall deadlock consists small cfs (#637).
+* fix conflicts between db_bench flags and enable speedb features flag(#743).
 * Proactive Flushes: Fix a race in the ShouldInitiateAnotherFlushMemOnly that may cause the method to return an incorrect answer (#758).
+* Stall deadlock consists small cfs (#637).
 
+* fix conflicts between db_bench flags and enable speedb features flag
 ### Miscellaneous
 * Remove leftover references to ROCKSDB_LITE (#755).
 * Options: Set compaction_readahead_size default to 0. The current default of 2Mb is not optimal for most of our use cases. Having a value of 0 means that the FS will use its default size for prefetching (true only with https://github.com/speedb-io/speedb/pull/788).
--- a/tools/db_bench_tool.cc
+++ b/tools/db_bench_tool.cc
@@ -577,7 +577,7 @@ DEFINE_bool(initiate_wbm_flushes,
 DEFINE_uint32(max_num_parallel_flushes,
               ROCKSDB_NAMESPACE::WriteBufferManager::FlushInitiationOptions::
                   kDfltMaxNumParallelFlushes,
-              "In case FLAGGS_initiate_wbm_flushes is true, this flag will "
+              "In case FLAGS_initiate_wbm_flushes is true, this flag will "
               "overwrite the default "
               "max number of parallel flushes.");
 
@@ -2009,6 +2009,9 @@ DEFINE_uint64(total_ram_size, 512 * 1024 * 1024ul,
               "SharedOptions total ram size bytes. ");
 namespace ROCKSDB_NAMESPACE {
 namespace {
+bool is_default(const char* flag_name) {
+  return gflags::GetCommandLineFlagInfoOrDie(flag_name).is_default;
+}
 static Status CreateMemTableRepFactory(
     std::shared_ptr<MemTableRepFactory>* factory) {
   Status s;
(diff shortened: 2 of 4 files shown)


tools/db_bench_tool.cc
```diff
@@ -9549,14 +9552,72 @@ void ValidateMetadataCacheOptions() {
   }
 }
 
-void ValidatePinningPolicyRelatedFlags() {
+void ValidateEnableSpeedbFlags() {
```
udi-speedb, line 9555: What about:
- pinning_policy
- filter_policy
- db_write_buffer_size
- bytes_per_sync

udi-speedb, line 9555: This is very verbose and error prone. In addition, it should probably be more or less duplicated in db_stress (anywhere else).
In view of that, I think we should strive to do the following:
- Have the validation code in some common location that may be used by both db_bench & db_stress
- Maybe have the validation use some list of flag names (strings). The code will go over the list and validate that all of the flags are in their default values. That will make the code shorter, somewhat more generic, and easier to maintain.

tools/db_bench_tool.cc
```diff
@@ -9549,14 +9552,72 @@ void ValidateMetadataCacheOptions() {
...
+          "together");
+    }
+    if (is_default("prefix_size") && FLAGS_memtablerep != "hash_spdb") {
+      ErrorExit(
+          "enable_speedb_features and memtablerep != hash_spdb cannot be "
+          "configured "
+          "together");
+    }
+  } else if (!is_default("total_ram_size")) {
+    ErrorExit("total_ram_size configured without enabling speedb features");
```
udi-speedb, line 9620: If use_blob_cache, blob_cache will be set to cache_. Is that OK?

tools/db_bench_tool.cc
```diff
@@ -9577,7 +9638,7 @@ int db_bench_tool_run_group(int group_num, int num_groups, int argc,
   parsing_cmd_line_args = false;
 
   ValidateAndProcessStatisticsFlags(first_group, config_options);
-  ValidatePinningPolicyRelatedFlags();
+  ValidateEnableSpeedbFlags();
```
udi-speedb, line 9641: An issue that needs consideration is related to the ValidateXXXX() methods.
They are using the flags to perform their validations (*e.g., ValidatePinningRelatedOptions()). However, the EnableSpeedbFeatures will run later and reconfigure the entities, regardless of the values of the flags. That means the validation may actually be meaningless.

udi-speedb:
What about HISTORY.md?

udi-speedb (changes requested):

tools/db_bench_tool.cc
```diff
@@ -9549,14 +9552,72 @@ void ValidateMetadataCacheOptions() {
...
+      ErrorExit(
+          "enable_speedb_features and use_dynamic_delay == false cannot be "
+          "configured together");
+    }
+    if (!is_default("cache_size")) {
+      ErrorExit(
+          "enable_speedb_features and cache_size cannot be configured "
+          "together");
+    }
+    if (FLAGS_cache_type != "lru_cache") {
```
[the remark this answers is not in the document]
ofriedma: let's keep it as !is_default

tools/db_bench_tool.cc
```diff
@@ -9549,14 +9552,72 @@ void ValidateMetadataCacheOptions() {
...
+      ErrorExit(
+          "enable_speedb_features and cache_type != lru_cache cannot be "
+          "configured together");
+    }
+    if (!is_default("simcache_size")) {
+      ErrorExit(
+          "enable_speedb_features and simcache_size cannot be configured "
+          "together");
+    }
+    if (FLAGS_prefix_size && FLAGS_memtablerep != "skip_list") {
```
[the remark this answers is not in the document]
ofriedma: will remove it

tools/db_bench_tool.cc
ofriedma: AFAIK blob cache is not being set with enable speedb features

udi-speedb: It is not. However, currently db_bench shares the single cache_ with blobs. With enable feature, it will use db_bench's cache_ while enable will define a separate cache to be used as the block cache

tools/db_bench_tool.cc
ofriedma: max_num_parallel_flushes , should be too?

HISTORY.md
```diff
@@ -10,6 +10,7 @@
 
 ### Bug Fixes
 * Stall deadlock consists small cfs (#637).
+* fix conflicts between db_bench flags and enable speedb features flag
```
udi-speedb, line 13: Please add the issue number and also place it at the end, to reflect the order in which items were actually added

tools/db_bench_tool.cc
```diff
@@ -9549,14 +9552,33 @@ void ValidateMetadataCacheOptions() {
...
+      "num_high_pri_threads", "num_bottom_pri_threads",
+      "num_bottom_pri_threads", "num_low_pri_threads",
+      "max_background_compactions", "max_background_flushes", "cache_size",
+      "cache_type",
+      // Assume simcache_size default is disabled simcache
+      "simcache_size", "memtablerep", "pinning_policy",
+      "scoped_pinning_capacity", "use_ribbon_filter", "bloom_bits",
+      "allow_wbm_stalls", "db_write_buffer_size", "initiate_wbm_flushes",
+      "bytes_per_sync", "use_dynamic_delay", "memtable_bloom_size_ratio",
+      "whole_key_filtering", "optimize_filters_for_hits"};
```
udi-speedb, line 9566: What about use_blob_cache?

tools/db_bench_tool.cc
udi-speedb: The short anser is yes.
Better include it as well. EnableSpeedbFeatures is not calculating it, but also not accepting it as a parameter to use from the user. So, if the user explicitly set it in db_bench, it will have no affect.
And please also include start_delay_percent for the same reasons.

udi-speedb (changes requested):

HISTORY.md
```diff
@@ -9,8 +9,9 @@
 * Added a kUseBaseAddress flag and GetBaseOffset flag to OptionTypeInfo.  If this flag is set and a function is used for processing options, the function is passed the base address of the struct rather than the specific field (#397)
 
 ### Bug Fixes
-* Stall deadlock consists small cfs (#637).
+* fix conflicts between db_bench flags and enable speedb features flag(#743).
```
udi-speedb, line 12: Please move to the end to reflect the order in which updates were made.

udi-speedb (approved):

udi-speedb:
@ofriedma - Before we merge, please ask @erez-speedb to test the branch, use db_bench, etc. Thanks

ofriedma:
@udi-speedb please review, no conflicts were after rebase

udi-speedb (approved):

ofriedma:
@udi-speedb please look at all conversation resolv them it is ok and tell me if it does not

PR #719, opened 2023-10-15
Use OptionsLogFormatter for Options::Dump

mrambacher:
This PR replaces the code in the Options::Dump method with code using the Options infrastructure.  A new "SerializePrintableOptions" method was added to the Configurable class which allows the values not serializable by the options to be included in the output.  A ConfigOptions::Depth::KPrintable was added to tell the classes to invoked this method as part of the serialization process.

The Options::Dump methods now use the this infrastructure to generate their output.  The code in the GetPrintableOptions methods of various was either deleted (because the values were serialized as Options) or moved into the SerializePrintableOptions method.

PR #770, opened 2023-11-19
Only dump the changed ColumnFamilyOptions to the LOG

### LINKED ISSUE #430: Options for only the first 10 column families are reported to the log
The options of column families are reported to the log at the top of every log file. However, if there are more than 10 column families (not very common but definitely allowed and occurs in practice), only the **options** of the first 10 are reported to the log.
Throughout the log file, any other log line that is associated with any column family will be reported. So, you find in the log information about column families whose options you don't know.

mrambacher:
This PR depends on #719 and closes #430.

This PR uses the recent changes to the Configurable code and the new OptionsFormatter class to only dump the changed options to the LOG file. Because the resultant ColumnFamilyOptions should be much smaller, the limit on the number of Column Families to dump has been removed.


ofriedma (changes requested):
LGTM, let's wait for #719

ofriedma (dismissed):

ofriedma:
was able to compile on arm

Issue #787, opened 2023-12-24
Deprecation of SetupForCompaction causes slow compactions


Yuval-Ariel:
**Describe the bug**
During the rebase on 8.6.7 (https://github.com/speedb-io/speedb/issues/736) we've encountered very long stalls as seen below:
![image](https://github.com/speedb-io/speedb/assets/105581454/35db56a5-f78f-4b37-996e-c0d0b9989161)
Where purple is the rebased branch and light blue is the current main. 

Further investigation showed that the stalls happened since the compaction ran much slower. Running iostat during that time showed that in the rebased version, the reads from disk are consistently at 4kb (rareq-sz) and no reads are queued (rrqm/s). 
We've tried playing with `compaction_readahead_size` as written in https://smalldatum.blogspot.com/2023/11/debugging-perf-changes-in-rocksdb-86-on.html but all values showed some degradation.

Release 2.7 of Speedb is based on RocksDB 8.1.1 and there, the compaction reads are using the filesystem defaults ('compaction_readahead_size' is 0 by default) for prefetching which results in much better performance than using any `compaction_readahead_size` value with the rebased version (rebased on Rocksdb 8.6.7). Some of this change was added in https://github.com/facebook/rocksdb/pull/11631. 

The reason for the slower performance of the compaction read speed in the rebased version when using `compaction_readahead_size = 0` (using the OS page cache) is because of this PR https://github.com/facebook/rocksdb/pull/11658, which removes hinting the FS with `POSIX_FADV_NORMAL` for files picked for compaction (which is the default value of `access_hint_on_compaction_start`). 
The removal of this Hint results in degradation since these files are already hinted with `POSIX_FADV_RANDOM` in `TableCache::GetTableReader` when the files are opened (controlled by flag `advise_random_on_open`, true by default).
 
RocksDB plan to deprecate the flag access_hint_on_compaction_start in release 9.0.

To fix the above issue, its been decided that files undergoing compaction will be hinted with `POSIX_FADV_NORMAL`.

Yuval-Ariel:
Performance is restored after reverting the code in SetupForCompaction (https://github.com/facebook/rocksdb/pull/11658)
Where red is the rebased branch with revert patch and green in current main.
![image](https://github.com/speedb-io/speedb/assets/105581454/dc6a8bd5-2acf-4be2-99e7-f81f80b4749a)

PR #788, opened 2023-12-25
Compaction: Restore SetupForCompaction functionality

Yuval-Ariel:
In https://github.com/facebook/rocksdb/pull/11658, files are no longer controlled by the access_hint_on_compaction_start flag by removing the content of SetupForCompaction. However, this leads to degradation in compaction read speed when combined with compaction_readahead_size since once a file is opened, in linux, it is hinted with POSIX_FADV_RANDOM.

See https://github.com/speedb-io/speedb/issues/787 for full details.

Add back the default hint to avoid degradation in this scenario.


ofriedma (dismissed):
LGTM

ofriedma (approved):

Issue #793, opened 2023-12-26
CI failed after changing default compaction_readahead_size from 2M to 0


ofriedma:
CI failed after changing default compaction_readahead_size from 2M to 0

PR #794, opened 2023-12-26
fix CI failure after changing compation_readahead_size default to 0

ofriedma:
fix CI failure after changing compation_readahead_size default to 0

fixes: https://github.com/speedb-io/speedb/issues/793


Yuval-Ariel (dismissed):

Yuval-Ariel (approved):

ofriedma:
@Yuval-Ariel Can we merge it?

PR #797, opened 2023-12-31
stress test: Add the ability to trace all operations

Yuval-Ariel:
Controlled by flag trace_ops (on by default).
Trace files will be written to the expected values dir


ofriedma:
@Yuval-Ariel Could it impact the correctness of the db_stress? (I guess that no)

Yuval-Ariel:
IT will not affect the correctness.

ofriedma (approved):
LGTM

PR #801, opened 2024-01-04
LOG Enhancement:Have a separate LOG entry per CF Stats

udi-speedb:


db/db_impl/db_impl.cc
```diff
@@ -1141,30 +1143,43 @@ void DBImpl::DumpStats() {
...
-                                                  &stats);
+                                                  &db_stats);
 
     property = &InternalStats::kPeriodicCFStats;
     property_info = GetPropertyInfo(*property);
     assert(property_info != nullptr);
     assert(!property_info->need_out_of_mutex);
     for (auto cfd : *versions_->GetColumnFamilySet()) {
       if (cfd->initialized()) {
+        std::string cf_stats_str{std::string("CF Stats [") + cfd->GetName() +
```
Yuval-Ariel, line 1154: why is this needed?
just below it theres a print which has the info about the cf name:
``` ** Compaction Stats [column_family_name_000018] ** ```

Comment (udi-speedb) [reply db/db_impl/db_impl.cc:1154]: Makes it more readable IMO

Comment (Yuval-Ariel) [reply db/db_impl/db_impl.cc:1154]: this is what it looks like:
``
2024/01/04-10:00:25.147816 69291 [/db_impl/db_impl.cc:1166] CF Stats [column_family_name_000024]
** Compaction Stats [column_family_name_000024] **
```
do you need the timestamp? because the upper line (the one you added) doesnt seem to add anything else

udi-speedb: What do you mean?
The whole point is separating each cf stats to a separate log entry. 
A log entry starts with a timestamp

Yuval-Ariel (dismissed):

Yuval-Ariel (approved):

PR #806, opened 2024-01-07
798 ci collect logs and save the db when qa test fails

maxb-io:
added 2 steps. the first will be executed when the workflow failed and create a tar file with the test db  and the second step will upload the tar as an artifact


speedbadmin (approved):

maxb-io (commented):
comment 01

Issue #813, opened 2024-01-23
stress test fail: Expected State - assertion IsDone failed


Yuval-Ariel:
```
db_stress: db_stress_tool/expected_state.cc:399: virtual rocksdb::{anonymous}::ExpectedStateTraceRecordHandler::~ExpectedStateTraceRecordHandler(): Assertion `IsDone()\' failed.
#7  0x00007f9a52aea729 in __assert_fail_base (fmt=0x7f9a52c80588 "%s%s%s:%u: %s%sAssertion `%s\' failed.\
%n", assertion=0x55f5bbb04740 "IsDone()", file=0x55f5bbb04600 "db_stress_tool/expected_state.cc", line=399, function=<optimized out>) at assert.c:92
92\tassert.c: No such file or directory.
#8  0x00007f9a52afbfd6 in __GI___assert_fail (assertion=assertion@entry=0x55f5bbb04740 "IsDone()", file=file@entry=0x55f5bbb04600 "db_stress_tool/expected_state.cc", line=line@entry=399, function=function@entry=0x55f5bbb046a0 "virtual rocksdb::{anonymous}::ExpectedStateTraceRecordHandler::~ExpectedStateTraceRecordHandler()") at assert.c:101
101\tin assert.c
#9  0x000055f5bb8abd48 in rocksdb::(anonymous namespace)::ExpectedStateTraceRecordHandler::~ExpectedStateTraceRecordHandler (this=<optimized out>, __in_chrg=<optimized out>) at db_stress_tool/expected_state.cc:399
399\t  ~ExpectedStateTraceRecordHandler() { assert(IsDone()); }
#10 0x000055f5bb8ce028 in rocksdb::(anonymous namespace)::ExpectedStateTraceRecordHandler::~ExpectedStateTraceRecordHandler (this=<optimized out>, __in_chrg=<optimized out>) at /usr/include/c++/9/ext/atomicity.h:69
69\t    return __result;
#11 rocksdb::(anonymous namespace)::ExpectedStateTraceRecordHandler::~ExpectedStateTraceRecordHandler (this=<optimized out>, __in_chrg=<optimized out>) at db_stress_tool/expected_state.cc:399
399\t  ~ExpectedStateTraceRecordHandler() { assert(IsDone()); }
[...]

PR #816, opened 2024-01-24
stress test: Disable IsDone assertion which fails with tracing sometimes

Yuval-Ariel:
This should be reverted once #813 is handled

Adding a trace file by default in #797 has revealed some incompatibilities between the trace file and several configurations (more details in #813).
Keep the trace file and remove the IsDone assertion.


udi-speedb (dismissed):

udi-speedb (approved):

udi-speedb (approved):

Yuval-Ariel:
@udi-speedb  rebased on main with zlib fix.

udi-speedb (dismissed):

udi-speedb (approved):

PR #831, opened 2024-02-08
WriteController: Remove redundant setting delay reports with single db

Yuval-Ariel:
The WC currently reports to the log whenever a new delay request is handled (even if its the same rate). 
These msgs are redundant in cases where the WC is only shared with 1 db and hence 1 logger since similar msgs are reported in ColumnFamilyData::RecalculateWriteStallConditions.

Currently it looks like :
```
2024/02/11-12:34:41.726676 7213 [WARN] [/write_controller.cc:153] WC setting delay of 178892456, client_id: 0x563f8a2e8f60, client rate: 178892456
2024/02/11-12:34:41.726684 7213 [WARN] [/column_family.cc:1235] [default] Stalling writes because we have 12 level-0 files rate 178892456
```
This PR removes the msg from the write_controller (the first) when there is only 1 logger to report to.

### DIFF
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -1,6 +1,16 @@
 # Speedb Change Log 
 
 ## Unreleased
+
+### New Features 
+
+### Enhancements
+
+### Bug Fixes
+
+### Miscellaneous
+
+## Incaberry 2.8.0 (31/1/2024)
 Based on RocksDB 8.6.7
 
 ### New Features 
@@ -38,7 +48,7 @@ RocksDB has a value of 10 by default and we've added the option to randomize the
 * Options: Set level_compaction_dynamic_level_bytes as false by default. This flag is not working properly with Speedb. see https://github.com/speedb-io/speedb/issues/786 for more details.
 * zlib: Update links to zlib 1.3 in CI and Makefile since the link in zlib.net is dead.
 
-## Hazlenut 2.7.0 (27/10/2023)
+## Hazelnut 2.7.0 (27/10/2023)
 Based on RocksDB 8.1.1
 
 ### New Features 
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -9,6 +9,7 @@
 ### Bug Fixes
 
 ### Miscellaneous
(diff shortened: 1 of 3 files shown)


udi-speedb:
@Yuval-Ariel - Why not avoid logging in the WC to the reporter's logger?
This would achieve what this PR is for as well as have a single log record (instead of 2) in the reporter's log.

Yuval-Ariel:
> @Yuval-Ariel - Why not avoid logging in the WC to the reporter's logger? This would achieve what this PR is for as well as have a single log record (instead of 2) in the reporter's log.

@udi-speedb , what you're suggesting wouldnt work when a logger is shared between two dbs. the db that initated the request will get a msg but the other one wouldnt.

udi-speedb:
> > @Yuval-Ariel - Why not avoid logging in the WC to the reporter's logger? This would achieve what this PR is for as well as have a single log record (instead of 2) in the reporter's log.
> 
> @udi-speedb , what you're suggesting wouldnt work when a logger is shared between two dbs. the db that initated the request will get a msg but the other one wouldnt.

Wouldn't the reporter get a log message from its own column_family.cc code while the other(s) would log the WC's message?

Yuval-Ariel:
> > > @Yuval-Ariel - Why not avoid logging in the WC to the reporter's logger? This would achieve what this PR is for as well as have a single log record (instead of 2) in the reporter's log.
> > 
> > 
> > @udi-speedb , what you're suggesting wouldnt work when a logger is shared between two dbs. the db that initated the request will get a msg but the other one wouldnt.
> 
> Wouldn't the reporter get a log message from its own column_family.cc code while the other(s) would log the WC's message?

its the same logger. so both dbs will get any msg

udi-speedb:
OK, but if there multiple delay sources (cf-s, WBM) wouldn't you miss the min rate and only see the new rate?

Yuval-Ariel:
> OK, but if there are multiple delay sources (cf-s, WBM) wouldn't you miss the min rate and only see the new rate?

we would still get the msg from the cf that initiated the delay

PR #861, opened 2024-06-12
db_bench: optimize Done, replace slow divisions with addition/subtrac…

ColinIanKing:
…tions

There are two slow 64 bit division operations that are computationally quite expensive. Replace these with a modulo style counter that uses simpler and faster increment, comparision and subtraction operations. On a 60 second run, this optimization reduced the function CPU utilization from 21.8 seconds down to 2.3 seconds on an i7-6700 as measured using Intel vtune.


ofriedma (approved):

ofriedma:
@ColinIanKing please add a record to HISTORY.md

